PR #209, opened 2020-07-10
fix: relative logo url with option to override

masone:
we hardcoded the logo url to https://www.carforyou.ch and in hindsight, I find that somewhat critical. I often find myself testing locally, clicking the logo and then not realizing I switched to production. I propose to default to `/`, while giving the private seller header in dealerhub the opportunity to override with `process.env.LISTINGS_URL` explicitly

### DIFF
--- a/src/components/header/index.tsx
+++ b/src/components/header/index.tsx
@@ -16,6 +16,7 @@ import useModal from "../../hooks/useModal"
 
 interface Props {
   theme?: HeaderTheme
+  logoUrl?: string
   mobileMenuText?: string
   links?: Array<JSX.Element>
   ctaButton?: JSX.Element
@@ -29,6 +30,7 @@ interface Props {
 
 const Header: FC<Props> = ({
   theme = HeaderTheme.LIGHT,
+  logoUrl = "/",
   links = [],
   ctaButton = null,
   iconButtons = [],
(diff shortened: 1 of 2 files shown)


lkappeler:
@Averethel Just for my understanding why do you see changing a link on a logo as a breaking change?

Averethel:
@lkappeler I looked if we need to change the usage. If we don't do anything in dealer hub the logo would link to the welcome page instead of the homepage of car for you. From that perspective, I see it as a breaking change since (if we don't do anything) the user workflow is broken. But it's also fair to not treat it as such since from the interface perspective it's an additive change. I'd like to keep one perspective when we look at changes though so we have the same reference.

masone:
yeah I was thinking about that too and decided to turn a blind eye to the fact that for a short time, dealerhub for users might link to dh.cfy.ch/ instead cfy.ch/

masone:
I'll bundle this with the separate repository we discussed. that's a clear breaking change anyway

lkappeler:
Yeah that was exactly my point like I'd always look the interface perspective of a package like. If you don't change anything when updating will it break your app or not...
(Ofc we need to update the link and add it when updating it but worst case the link we point to another target but nothing will be broken in that way...)

PR #210, opened 2020-07-10
feat: tooltip component

Averethel:
Adds a tooltip component:
Screenshot 2020-07-10 16 03 26

todo:
- [x] tooltip component
- [x] verify mobile behaviour
- [x] switch tooltip side if there's not enough space

@carforyou/frontend reviews are welcome, there might be some minor changes coming from behaviour on mobile that I need to verify


Averethel:
@carforyou/frontend I'm pretty much done with the implementation here. Reviews would be nice.

src/components/tooltip/popUp.tsx
```diff
@@ -0,0 +1,45 @@
...
+        /* eslint-disable @typescript-eslint/naming-convention */
+        "tooltip_topLeft bottom-tooltip mb-10 right-0":
+          position === TooltipPosition.topLeft,
+        "tooltip_topRight bottom-tooltip mb-10 left-0":
+          position === TooltipPosition.topRight,
+        "tooltip_bottomLeft top-tooltip mt-10 right-0":
+          position === TooltipPosition.bottomLeft,
+        "tooltip_bottomRight top-tooltip mt-10 left-0":
+          position === TooltipPosition.bottomRight,
+        /* eslint-enable @typescript-eslint/naming-convention */
```
ufkaya, line 37: You can avoid so many conditions by moving the classNames to the tooltip.css

Averethel: Yes and no. I would need to whitelist classes if I'd construct them like `tooltip_${position}`.
Also, I thought we only use standalone CSS if we can't achieve things with a tailwind class...

Averethel:
@carforyou/frontend since the tooltip position switch calculation needs to happen when showing the tooltip I needed to add it here instead of the consumer. Can I have another review?

src/components/tooltip/index.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { FC, ReactNode, useState, useRef } from "react"
+
+import PopUp from "./popUp"
+
+export enum TooltipPosition {
```
lkappeler, line 5: To me, it's very new to have a tooltip that defines more that 4 positions, TRBL...

src/components/tooltip/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  bottom = "bottom",
+  bottomLeft = "bottomLeft",
+  bottomRight = "bottomRight",
+  left = "left",
+  right = "right",
+}
+
+const spacingPosition = {
+  top: "top",
+  topLeft: "top",
```
lkappeler, line 18: That's an indicator on the first comment I made about the position options Like at first I did not understand what the intention is behind this...

src/components/tooltip/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  topRight: "top",
+  bottom: "bottom",
+  bottomLeft: "bottom",
+  bottomRight: "bottom",
+  left: "left",
+  right: "right",
+}
+
+const oppositePosition = {
```
lkappeler, line 27: To me, it feels very counter-intuitive, probably from working with other tooltip libraries to switch directions 180deg when not enough space is available id rather expect them to top and then bottom as there, in theory, must be space in like 99isch per cent of the cases whereas left or right will be to small way more often if the page is to small

src/components/tooltip/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  bottom: window.innerHeight - bottom,
+  right: window.innerWidth - right,
+})
+
+export const calculateTooltipPosition = (
+  container,
+  originalPosition,
+  switchThreshold
+) => {
+  if (!switchThreshold && !container) {
```
lkappeler, line 50: Can you explain to me the concept with the switching threshold? Like I expect that to be something internal to the tooltip component and not or only with something like a build config configurable to the user. Like the tooltip should try to take another position on itself when there is not enough space...
In the storybook, I saw even in the threshold options. Some very strange options I'd never expect from a tooltip to behave like that. Like it switched sides and broke the page width. Id, in general, expects to take less space and line break the content. (Similar to explained above switching left to right)

src/components/tooltip/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+   * Threshold of the position switch
+   * If the space around tooltip container (on the side where tooltip is to be displayed)
+   * is smaller than then threshold the tooltip will be moved
+   */
+  positionSwitchThreshold?: number
+}
+
+const Tooltip: FC<Props> = ({
+  children,
+  renderContent,
```
lkappeler, line 82: What do we gain from a render props function in this case when we don't pass any functions into it like we could straight up replace it even without a custom hook and it would do the same?
As from [an article](https://kentcdodds.com/blog/when-to-not-use-render-props) regarding render props you sent me a while ago
> UPDATE: If you can use react@16.8.0 then the answer is to rarely use render props and almost always use a custom hook. HOOKS ARE ALMOST ALWAYS THE SUPERIOR PATTERN. Read about when you might still use render props

assets/tooltip.css
```diff
@@ -0,0 +1,63 @@
+.tooltip_top::after {
+  content: " ";
+  margin-left: -10px;
+  border-width: 10px;
```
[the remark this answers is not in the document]
Averethel: No, it's to render the tooltip arrow.

src/components/tooltip/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Position of the tooltip
+   */
+  position: TooltipPosition
+  /**
+   * Threshold of the position switch
+   * If the space around tooltip container (on the side where tooltip is to be displayed)
+   * is smaller than then threshold the tooltip will be moved
+   */
+  positionSwitchThreshold?: number
```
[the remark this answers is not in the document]
Averethel: In pixels, will add that to the docs.

src/components/tooltip/index.tsx
Averethel: Fair, it's about alignment to the parent element. Given the confusion, it might be a good idea to make it a separate prop, WDYT?

I need the variation since there's not enough space on mobile to show the tooltip centred above the parent.

src/components/tooltip/index.tsx
Averethel: My idea was that this is supposed to solve the case when the content is scrolled too far to accommodate for tooltip.
I want to limit the logic of checking the available space to this case and make the consumer handle responsive cases by adjusting the position accordingly. Does that make sense?

src/components/tooltip/index.tsx
Averethel: Yeah, I agree it's a bit clunky. Do you have a suggestion on how to check the size of the content that is not rendered though? Since the tooltip content is not in the DOM I can't just get the size of it and use it as a value...

src/components/tooltip/index.tsx
Averethel: I want to pass a tooltip content and not render it. And since I don't want to have a very awkward markup (like the library you linked) I went for render prop. That's one of the cases that can't be covered by a hook.

Averethel:
> 1. Do we really wanna implement and maintain ower own logic to handle tooltips, there are a ton of libraries out handling tooltip behaviour, which in my opinion, cover a ton of cases we gonna need a lot of time to implement ourselves. (A example would be [react-tooltip](https://wwayne.github.io/react-tooltip/) it might be a bit on the big side and we wanna evaluate a simpler library to not bundle so much for every page that has a tooltip) Personally I'd rather evaluate a library that serves our current needs and provide a wrapper component that serves the CFY layout with the required interface exposed to the user or the components library.

I'm pro pulling libraries that:
- solve complex components (like carousels or dropdowns)
- don't impose markup (like downshift)

A tooltip doesn't seem like something of a complexity that warrants an external library, especially for the limited use cases that we have. So yeah, I'd rather build something we can use than pull a library that bloats bundle size, enforces a markup and styling and we need to spend time customizing anyway.

> 2. If we really decide to custom implement tooltip logic:
>    2.1 Id simplify the interface to TRBL

How do you suggest to handle the case when tooltip cannot be centred over the parent element because there is not enough space? Would you use a separate property for that?

>    2.2 Make sure it never makes the page wilder than the original page

I don't fully agree that this should be the responsibility of the tooltip component. Whoever renders the tooltip should be able to ensure that.

>    2.3 Switch the behaviour form trying the 180deg position to probably try bottom and then try top, for the reason that if there is not enough space on the left/right changes are way higher bottom and if not (on the bottom of the page) top will be more space available to render the tooltip.

As you said that covers 99% of the cases. I don't feel like complicating logic to fix that. If it's the responsibility of the consumer to render tooltip we can live with the present switch which handles content scrolling.

An alternative would be to not do anything but allow passing position calculation as a function, this way the consumer could fully control the behaviour. WDYT?

Averethel:
Yeah the more I think about it the more I like a interface like:
```
   getPosition - ({ top, left, bottom, right }) => Tooltip position
```
With the argument being spacing around the tooltip container at the time of showing tooltip. Then we don't bake anything in and the consumer has all the control over the position change logic should it be necessary at all. We could pass tooltip dimensions (if we can calculate them in a sensible way) as a second argument too.

WDYT?

Averethel:
To sum-up changes discussed face to face:
let's switch to the following interface:
```
<Tooltip
   renderContent={() => <>Lorem ipsum</>}
   alignment={"start" | "center' | "end"}
   getPosition={({ top, left, right, bottom } - spacing around the tooltip container) => returns "top" | "left' | "bottom" | "right"}
 >I have a tooltip</Tooltip>
```

Averethel:
@lkappeler we have that documented in the description of the prop, which should be visible in the storybook.

PR #213, opened 2020-07-14
feat: add pill component

masone:
I will be using this in the header. the border-8 was not in the config and I want the header to use only tailwind values from components lib if possible

Screenshot 2020-07-14 at 19 50 21

### DIFF
--- a/src/components/pill.tsx
+++ b/src/components/pill.tsx
@@ -3,7 +3,7 @@ import classnames from "classnames"
 
 interface Props {
   children: ReactNode
-  secondary: boolean
+  secondary?: boolean
 }
 
 export const Pill: FC<Props> = ({ children, secondary }) => (
--- a/src/components/pill.tsx
+++ b/src/components/pill.tsx
@@ -1,0 +1,20 @@
+import React, { FC, ReactNode } from "react"
+import classnames from "classnames"
+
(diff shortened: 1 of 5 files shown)


src/components/pill.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { FC, ReactNode } from "react"
+import classnames from "classnames"
+
+interface Props {
+  children: ReactNode
+  secondary?: boolean
+}
+
+export const Pill: FC<Props> = ({ children, secondary }) => (
```
ufkaya, line 9: fyi the correct name is `badge` for this kind of component

masone: thanks! motivated by your comment, I aligned with Enrique on the naming. he thinks `pill` is appropriate. bootstrap calls these `pill badges`, so we're both kinda close :) https://getbootstrap.com/docs/4.0/components/badge/#pill-badges

I'd argue that `badge` is reserved for dealer badges, mbg badge, etc already

lkappeler:
Very nice! For the optics, you could add a div with some padding around the pill in the SB so it aligns with the description

masone:
fair input @lkappeler - it's the same for other components. I'll leave this out for now as a general separate improvement

PR #214, opened 2020-07-20
feat: onClose function for modals

lkappeler:
Adds the ability to pass a onClose callback to a modal

### DIFF
--- a/src/components/modal/index.tsx
+++ b/src/components/modal/index.tsx
@@ -16,10 +16,11 @@ interface Props {
   style: ModalStyle
 }
 
-const Modal: FC<Props> = ({ close, size, style, children }) => {
+const Modal: FC<Props> = ({ close, onClose, size, style, children }) => {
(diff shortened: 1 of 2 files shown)


src/components/modal/index.tsx
```diff
@@ -10,15 +10,17 @@ export type ModalStyle = "white" | "dark"
 
 interface Props {
   close: () => void
+  onClose?: () => void
```
masone, line 13: couln't you add code to the close function instead?

```
<Modal close={() => close(); afterClose(); }
```

masone: ah, that would not cover the case where we rely on the "internal close"

lkappeler: Wouldn't that break the interface of the component and we would need to update it everywhere over two projects?

lkappeler:
@masone Will create a PR with some testis and update the story book once I finished my open taksk :)

PR #216, opened 2020-07-22
fix: Modal onClose behaviour

lkappeler:
- fix: calls the onClose event when the user clicks on the overlay
- tests: unit tests for all onClose cases
- story book: adds example for the onClose event

Screenshot 2020-07-22 at 14 17 30

### DIFF
--- a/src/hooks/__tests__/useModal.test.tsx
+++ b/src/hooks/__tests__/useModal.test.tsx
@@ -69,46 +69,39 @@ describe("useModal", () => {
       fireEvent.keyDown(container, { keyCode: 27 })
 
       expect(container.firstChild).toMatchSnapshot()
+      expect(container).not.toContain(modalText)
     })
 
-    it.only("calls the onClose function when pressing ESC", async () => {
+    it("calls the onClose function when pressing ESC", async () => {
       const onCloseMock = jest.fn()
-      const { getByText, container } = renderWrapper(onCloseMock)
+      const { getByText, container, getByTestId } = renderWrapper(onCloseMock)
 
-      await act(async () => {
-        await fireEvent.click(getByText(buttonText))
-        getByText(modalText)
-        await fireEvent.keyDown(container, {
-          key: "ESC",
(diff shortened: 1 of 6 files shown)


stories/modal/index.stories.tsx
```diff
@@ -1,10 +1,22 @@
...
-  .add("Large", generateStory({ size: "large" }))
-  .add("Fulscreen", generateStory({ size: "fullscreen" }))
-  .add("Dark Fulscreen", generateStory({ style: "dark", size: "fullscreen" }))
+  .add("Small", generateModalStory({ size: "small" }))
+  .add(
+    "Small with onClose event",
+    generateModalStory({
+      size: "small",
+      onClose: () => {
+        alert("onClose callback called")
```
Averethel, line 12: Could you use `action` from [actions add on](https://www.npmjs.com/package/@storybook/addon-actions) instead? It will then pop-up in the action tab.

Check out how it's used in in the button story: https://github.com/carforyou/carforyou-components-pkg/blob/master/stories/button.stories.js#L9

lkappeler: Nice wasn't aware of it though like what's the most recognizable way to showcase the callback was executed 😆

src/hooks/__tests__/useModal.test.tsx
```diff
@@ -66,6 +69,41 @@ describe("useModal", () => {
...
+    })
+
+    it("calls the onClose function when the overlay is clicked", async () => {
+      const onCloseMock = jest.fn()
+      const { getByText, getByTestId } = renderWrapper(onCloseMock)
+      fireEvent.click(getByText(buttonText))
+      getByText(modalText)
+      fireEvent.click(getByTestId("modal-overlay"))
+
+      expect(onCloseMock).toBeCalledTimes(1)
```
masone, line 106: perfect, thanks!

lkappeler:
@Averethel Switched to the action

PR #218, opened 2020-07-23
fix: Modal onClose behaviour

lkappeler:
Github seems to have made some changes, I don't get an edit filed any more when merge the PR and it picks the latest commit message instead of the title, therefore, https://github.com/carforyou/carforyou-components-pkg/pull/216 was not released 😕


lkappeler:
My bad somehow message up the squash merge

PR #220, opened 2020-07-29
feat: tooltip revamp

Averethel:
We had a pretty nice and simple implementation of a tooltip component.
Unfortunately, it had the same problem any simple implementation of tooltip has - the dreaded `z-index`.

Normally it wasn't a problem but in the case of carousel cards we ended up with the following:
Screenshot 2020-07-28 15 24 16

The reason for it is `z-index` stacking and all `flex` containers creating their own stacking context. In essence, our carousels case is very similar to the case described here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/Stacking_context_example_3

After struggling a lot trying to make it work with carousels and everything I decided to revisit using a library for tooltips. 
Using the library also takes away the pain of positioning tooltip if there's not enough space which allows us to simplify the interface (hence it will be a **BREAKING CHANGE**)

Here's a preview of how the same problematic tooltip looks when using the new implementation:
Screenshot 2020-07-29 12 58 30


assets/tooltip.css
```diff
@@ -1,95 +1,66 @@
...
-  border-width: 10px;
-  bottom: 10px;
-  border-color: transparent white transparent transparent;
-  @apply border-solid absolute right-full;
+.tooltipArrow-right::after {
+  @apply top-0;
+  border-color: transparent transparent transparent white;
+  border-width: 10px 0 10px 10px;
+  left: -5px;
 }
```
masone, line 66: can you quickly state what's holding us back from doing these with tailwind? this adds to the global CSS bundle while we have some parallel initiatives to reduce that size. it turns out to be pretty hard to build it back to tailwind once we have custom CSS

package.json
```diff
@@ -34,6 +34,7 @@
     "downshift": "^5.0.0",
     "glob": "^7.1.6",
     "react-paginate": "^6.3.2",
+    "react-popper-tooltip": "^3.1.0",
```
masone, line 37: there are alternatives that are 80% smaller. are you absolutely positively sure that it's worth adding 30kb to the bundle for this?

stories/tooltip/generator.tsx
```diff
@@ -32,7 +32,7 @@ const generateStoryFunction = ({ position, alignment }) => () => {
               gravida.
             </p>
           )}
-          getPosition={() => position}
+          preferredPosition={position}
```
masone, line 35: why is it `preferred` and not just the absolute position? will it pop down when there's no space to pop up even though I preferred it to go up?

assets/tooltip.css
Averethel: As far as I know, it's not possible to style `after` pseudo-elements with tailwind.

package.json
Averethel: Most of the very small alternatives that I've seen are naive implementations based on `z-index` (similar to what we have before). I believe it's well worth it to use a well-established engine here and not worry about tooltips anymore.

stories/tooltip/generator.tsx
Averethel: As the docs state: https://github.com/carforyou/carforyou-components-pkg/pull/220/files#diff-4f6c0f244f5e6da59ac54389842bea08R32-R34

So yes, you're correct - if we would want to render tooltip on the top but there's not enough space it will be rendered on the bottom instead.

package.json
Averethel: That being said - if you know of a small alternative that uses non-z-index approach I'd be more than happy to give it a shot. I looked at quite some libraries but might've missed something.

stories/tooltip/generator.tsx
Averethel: I opted for adding `preferred` prefix to signify that. Happy to switch back to `position` like most of the tooltips do

package.json
lkappeler: Yeah I'm very ambivalent about this, from my experience I do see the problem when electing, small libraries as they tend to not hold up for every case, on the other side increasing the package by [1/3](https://bundlephobia.com/result?p=@carforyou/components@9.0.0-tooltip-revamp-1d964f19a517c897b5d5d4c58fba996c5771af37.1) for a rather hidden feature is quite massive on the other side. So

assets/tooltip.css
lkappeler: It's a bit hard to see in the diff but would it be possible to only keep the before/after classes in the global CSS to minimize it and use tailwind for the non-pseudo classes?

package.json
Averethel: I see your concern. On the other hand, this enables us to build complex components like date pickers or multi-level menus so I think it'll pay off in the long run. Plus we're still being tree shakeable so you'd only include the extra js if the tooltips are rendered, no?

assets/tooltip.css
Averethel: My intention was to not add extra values to tailwind config since they are only used for the tooltip. Happy to change that if we prefer tailwind.

Averethel:
@masone @lkappeler I am more than happy to hear suggestions for an alternative library. I did not manage to find something else that works well and does not impose markup and or styling, but it could be that I've missed something - if you have some specific alternative in mind please let me know.

Averethel:
@masone @lkappeler I did another round of library checks. Somewhere in the depths of the GitHub I foundhttps://github.com/bsidelinger912/react-tooltip-lite. It has a smaller footprint than the popper variant.
I think this is the smaller we can go without going with something singleton and data-attributes based or developing a complex positioning engine ourselves.

It increases from 78kB to 93.8kB (minified)/22.3kB to 25.9kB (minified + gzipped) which is considerably smaller than the popper approach

assets/index.css
```diff
@@ -5,4 +5,3 @@
 @import "./spinner.css";
 @import "./intercom.css";
 @import "./header.css";
-@import "./tooltip.css";
```
masone, line 8: likey likey

src/components/tooltip/index.tsx
```diff
@@ -1,46 +1,36 @@
...
-
-import PopUp from "./popUp"
+import React, { FC, ReactNode } from "react"
+import TooltipTrigger from "react-tooltip-lite"
 
 export enum TooltipPosition {
-  top = "top",
-  bottom = "bottom",
+  up = "up",
+  down = "down",
```
masone, line 6: don't forget to indicate a  BREAKING CHANGE :)

src/components/tooltip/index.tsx
```diff
@@ -50,38 +40,22 @@ interface Props {
...
-      }}
-      className="relative inline-block"
-    >
-      {isVisible ? (
-        <PopUp position={calculatedPosition} alignment={alignment}>
-          {renderContent()}
-        </PopUp>
-      ) : null}
-      {children}
+    <div className="inline-block" onClick={(e) => e.preventDefault()}>
```
masone, line 47: will links within the tooltip work with this?

Averethel: Yes, it's to enable having tooltips inside links (card)

Averethel:
@lkappeler The space in storybook comes from the previous bugfix when we removed hardcoded tooltip width. I took a random defined width class and it seems to not be enough space for long words to fit. I'm not too concerned here

PR #224, opened 2020-08-10
feat: add fixed card

ralucacril:
Extracted the fixedCard component
Screenshot 2020-08-11 at 17 05 59

I had to add a new folder - public - for digesting images in storybook. If there is a different way of doing that, let me know. Thanks!


stories/card/fixed.stories.js
```diff
@@ -0,0 +1,31 @@
...
+    Description
+    ~~~~
+    <FixedCard />
+    ~~~
+    `)(() => (
+    <div className="mx-30 mb-40">
+      <div className="text-2xl mb-20">Example</div>
+      <div className="w-3/12">
+        <FixedCard renderImage={() => <img src="/dummyImage.jpeg" />}>
+          <span className="pl-16 py-15">
```
lkappeler, line 19: Is the padding something we could move to the component?
(If so we probably wanna use px instead of pl)

assets/card.css
```diff
@@ -0,0 +1,9 @@
+.fixedCard {
+  @apply flex flex-wrap rounded-10;
```
lkappeler, line 2: According to abstract the radius is 12px

image
https://app.abstract.com/projects/0935fec8-eab2-420e-b519-c7c185935030/branches/f98ccf2a-d63b-42fb-9f13-64de4a7272e5/commits/a45d120cfe8c08ba5f6ae18b59ac4b454d611b8d/files/27b5d2a6-1049-4b8c-91d2-026563d2d288/layers/9DED07A7-FCF3-49BD-AB83-0AD868F2112D?collectionId=19c9b4c0-fc5b-4260-9413-c98d4c7317d1&collectionLayerId=43f3f259-4973-4260-a18e-a03ccafc16f6&mode=build&selected=3076742694-09BC476B-E998-4AC2-A867-F984AE49584A

src/components/card/fixed.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { FC, ReactNode } from "react"
+
+interface Props {
+  renderImage: () => ReactNode
+  children: ReactNode
```
lkappeler, line 5: I think you don't need to type children in the props of a FC as they are provided by the FC interface

assets/card.css
GutierrezGo: The cards have `rounded-10`.
So, for consistency we leave it with 10px for all the cards.

GutierrezGo: And I have checked with Enrique and he agrees with 10px for all the cards.

stories/card/fixed.stories.js
ralucacril: umm....i would not do that since the component is just about the component and that span was just an example

ralucacril:
@lkappeler  the corners of the image are part of the image in the project. so is part of that, that is why i can't do anything about it.

src/components/card/fixed.tsx
ralucacril: i will check, thanks for letting me know

stories/card/fixed.stories.js
ralucacril: the padding is not part of the component.. but good eye 😄

src/components/card/fixed.tsx
ralucacril: i see that all the other components have the children as a prop...so i thought we go with the same pattern. WDYT?

stories/card/fixed.stories.js
lkappeler: Yeah i though like maybe we every card should have the same base padding for its text content to be consisten.....

src/components/card/fixed.tsx
lkappeler: No strong oppion its just not needed up to you

lkappeler:
@ralucacril What about just adding overflow hidden to the card?

image

stories/card/fixed.stories.js
Averethel: I'd like having padding as part of a component here if that's feasible. Would help us with consistent look & feel.

src/components/card/fixed.tsx
Averethel: We do require children here, whereas they are optional in FC type. I'd keep it.

Averethel:
@lkappeler IIRC it was on the card and then we moved it to an image container for some reason. What we could do is to add a container around the image within the card component though.

stories/card/fixed.stories.js
ralucacril: ok, then i will take a look

ralucacril: even though i am not sure i can due to the fact that we have the image inside of the component and that will break the whole component

ralucacril: so...if i add a wrapper around the children with the padding, and that children has a link, the clickable area won't be full width.

ralucacril:
@lkappeler and @Averethel added the rounded corners.

stories/card/fixed.stories.js
Averethel: We had a similar issue in `Button` and `Dropdown` we solved that with link wrapping - this could help: https://github.com/carforyou/carforyou-components-pkg/blob/master/src/lib/buttonHelper.tsx#L97

assets/card.css
```diff
@@ -0,0 +1,9 @@
...
+  @apply shadow-small bg-white relative z-card;
+  @apply flex-col overflow-hidden;
```
Averethel, line 4: I recall `overflow-hidden` on the whole card was problematic with some cases of the layout. Can we add a container with the overflow just around the image?

ralucacril: why would we add a container if this works? i would be more curious why you think is not working. maybe we needed to add the rounded corners on the images due to the fact we were having a slider instead of an image or another reason, which i can't think about now. I would not add an extra wrapper just because it might break. if is breaking, sure, i would add it, but otherwise i think is not neccesarry.

stories/card/fixed.stories.js
ralucacril: padding are different: growing card has completly different paddings than the bank card. so for that case having padding in the card is not a feasible solution. and i would like to avoid having the fixed card with paddings and the growing card without paddings.

assets/card.css
Averethel: If I recall correctly it messed up the card footer in the carousel cards.

Averethel: Please check if we have it directly on the fixed card in listings. I remember spending some time on putting overflows elsewhere than in card root, but it could be that my memory is tricking me.

ralucacril: we don not have it directly on the fixed card in listings. but i will check for you.

ralucacril: here 
Screenshot 2020-08-12 at 12 48 40

stories/card/fixed.stories.js
lkappeler: I see you point on not adding the padding in a fixed card cuz there is now default padding in the growing card, you could also add it there. (Not that deep into the differences between the card types...)
If form a desing perspective a different padding is needed on the types of fixed cards you could either add options for the padding as in "small" | "large" or take the smaller of both paddins and use it as default and add the additional padding for the option needed

As a side note you could also go with a link function to avoid passing classed as in , this would also remove the need of rendering the link twice in for the listing cards.
```
const FixedCard = ({ renderImage, children, onClick, renderLink }) => {
  const renderComponent = () =>(
    <article className="fixedCard" onClick={onClick}>
      {renderImage()}
      {children}
    </article>
  )

  return renderLink ? renderLink(renderComponent) : renderComponent()
}
```

ralucacril: i propose the following. i release with the current state, and then next sprint i will try to align with Enrique about the padding and then if they are the same for all the cards, i will make the changes or introduce small and big. WDYT?

assets/card.css
Averethel: Ah snap, nvm. It wasn't the footer. It was our initial tooltip implementation that was problematic. That's not the case anymore I think.

PR #226, opened 2020-08-13
feat: add growing card

ralucacril:
Screenshot 2020-08-13 at 13 07 54
Screenshot 2020-08-13 at 13 07 46

### DIFF
--- a/src/components/card/growing.tsx
+++ b/src/components/card/growing.tsx
@@ -1,0 +1,20 @@
+import React, { FC, ReactNode } from "react"
+import classNames from "classnames"
+
+interface Props {
+  renderImage: () => ReactNode
+  children: ReactNode
+  noMarginsOnMobile?: boolean
+}
+
+export const GrowingCard: FC<Props> = ({ renderImage, children, noMarginsOnMobile }) => {
+  return (
+    <article className={classNames("growingCard"
+    {"rounded-none md:rounded-10": noMarginsOnMobile})}>
+      {renderImage()}
+      {children}
+    </article>
+  )
+}
+
+export default GrowingCard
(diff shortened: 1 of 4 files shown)


src/components/card/growing.tsx
```diff
@@ -0,0 +1,27 @@
...
+interface Props {
+  renderImage: () => ReactNode
+  children: ReactNode
+  noMarginsOnMobile?: boolean
+}
+
+export const GrowingCard: FC<Props> = ({
+  renderImage,
+  children,
+  noMarginsOnMobile,
```
lkappeler, line 13: It's not really related to margins, isn't it? More like `roundedCornersOnMobile`

masone:
as discussed - we don't think that we can leverage this to make the dealer cards happen. ergo, this would be only used in listings. I would recommend not extracting it in that case to spare us from unnecessary maintenance overhead. only extract it if you see that we're sharing this component across projects

ralucacril:
@masone yap, i will not extract this.

src/components/card/growing.tsx
masone: I did not catch that, but now that you point at it - the name here seems overly specific. can we generalize it more? what would be a "semantic" approach? we have for example a `fullWidth` property on the `AlertMessage` https://github.com/carforyou/carforyou-components-pkg/blob/136ab6c03c84012f37c24568b09a566f80904819/src/components/alertMessage.tsx#L38

if `fullWidth` -> non-rounded corners
else -> rounded corners

masone: oh my. we closed this PR already 🙈

PR #231, opened 2020-08-20
feat: remove header

Averethel:
The header was moved to [its own package](https://github.com/carforyou/carforyou-header-footer-pkg), with all projects migrated we can remove it from components


GutierrezGo (commented):
What about the header.css?

Averethel:
@GutierrezGo removed now, good point 👍

GutierrezGo (commented):
Maybe the header values in tailwind.
They are 7 values.

GutierrezGo:
> @GutierrezGo That's awkward but we still use those in the header package and without them, the header would be b0rked.
> They would eventually go away once we're ready to switch to CSS modules in packages, which @masone is looking into

I see, thank you for making clear this point.

GutierrezGo (approved):

masone (approved):

PR #235, opened 2020-08-26
fix: use css modules for intercom

masone:
https://autoricardo.atlassian.net/browse/CAR-4713
this is a demo implementation of how to use css modules in the pkg.

- [ ] reorganize commits
- [ ] issue a breaking change
- [x] get storybook to work with css modules


Averethel (approved):
Whooop, whoop! That looks awesome 👍

lkappeler (approved):
Nice effort!

PR #238, opened 2020-08-31
CAR-4911: add openingHours component

GutierrezGo:
> Jira https://autoricardo.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=CAR&modal=detail&selectedIssue=CAR-4911

The design in the dealer page has changed. Therefore we need to adapt the clock icon.

Screenshot 2020-09-03 at 11 29 55

**Without clock icon in the opening hours component**
Screenshot 2020-09-03 at 11 31 11

**With clock icon in the opening hours component on PDP, next to the contact form (dealers)**

Screenshot 2020-09-03 at 11 33 55

You can see the component migrated here -> https://github.com/carforyou/carforyou-listings-web/pull/2357

**TO DO**
- [x] Check with Enrique which text-size we want in the opening hours. Because we have in listings `14px` and in DH `16px`. We need to make consistent the component.

So, after checking with Enrique, the text-size on the opening hours is `14px`

## Thank you 🌵


src/components/openingHours.tsx
```diff
@@ -0,0 +1,129 @@
+import React, { FC } from "react"
+
+import ClockOutlinedTeal from "./icons/clockOutlinedTeal"
+
+interface Props {
+  /**
+   * Any opening hours to be displayed
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  openingHours: any
```
masone, line 10: please don't type this as `any` but replicate the type from the api client

masone (approved):
thanks, looks good to me! I would like to specify a proper type for the openinghours. please add one before moving on

GutierrezGo:
@masone 
Should we leave the children as it is or should we remove it and add the disclaimer after the `openingHours` component?

I'm still in doubt about this point.

masone:
I feel rendering as a child does not add any value over rendering it after the component. so me personally, I would not render children

src/components/icons/clockOutlinedTeal.tsx
```diff
@@ -0,0 +1,20 @@
...
+        fillRule="evenodd"
+      >
+        <rect x={6} y={6} width={20} height={20} rx={10} />
+        <path strokeLinecap="round" d="M16 11.2V16l2.4 2.4" />
+      </g>
+    </svg>
+  )
+}
+
+export default ClockOutlinedTeal
```
masone, line 20: oh! this should be a png and be converted to a component automatically instead of being placed manually here - or am I wrong? I can't fully recall how the setup is in components-pkg

src/components/openingHours.tsx
```diff
@@ -0,0 +1,131 @@
...
+import ClockOutlinedTeal from "./icons/clockOutlinedTeal"
+
+interface Props {
+  /**
+   * Any opening hours to be displayed
+   */
+  openingHours: Array<{
+    open: { dayOfWeek: string; time: string }
+    close: { dayOfWeek: string; time: string }
+  }>
```
masone, line 12: 👍

masone (approved):
tiny question on the icon component - can't recall how they're handled here. but it's likely good to merge!

src/components/icons/clockOutlinedTeal.tsx
Averethel: No, we don't have that set up in components yet. We do have a tech-20 about icons in components in general, I'd tackle them together

GutierrezGo: Mmmmm good question, I will check it.
Thank you 🍦

GutierrezGo: thank you for the clarification @Averethel 🧸

Averethel (approved):

PR #258, opened 2020-11-11
feat: Add badges as components

gzimh:
Refs
- https://autoricardo.atlassian.net/browse/CAR-7235
- Buyer Protection https://autoricardo.atlassian.net/browse/CFYPD-43

Extracts the existing badges and adds the new buyer protection badge. Badges come with a common interface, the gdbd badge has an additional score variation. Small badges show only the logo.

```
export interface ExternalBadgeProps {
  language: "de" | "fr" | "it | "en"
  size?: "small" | "large"
  tooltipContent?: ReactNode
}
```

Default variations preview, check out the storybook to see all the available variations

Verified:
Screenshot 2021-06-02 at 13 48 22

MoneBackGuarantee:
Screenshot 2021-06-02 at 13 48 17

Good Deal Bad Deal
image
Screenshot 2021-06-02 at 13 47 56
image
Screenshot 2021-06-02 at 13 47 52
Screenshot 2021-06-02 at 13 47 48

BuyerProtection:
Screenshot 2021-06-02 at 13 47 41

BuyOnline:
Screenshot 2021-06-02 at 13 47 35

### How to test

Check out the branch and run `npm run dev`, check the storybook


ufkaya:
What is the status of this PR? Do we still need it?

djanov:
> What is the status of this PR? Do we still need it?

I think yes, we have now verified badge too, Colin already made a ticket on JIRA [CAR-7235](https://autoricardo.atlassian.net/browse/CAR-7235)

mreljin:
I'm guessing this is a bit outdated since storybook got updated since then. I need to add the verified ribbon and badge for my PR, should I just create a new PR and ticket for only those components?

src/tailwind/defaultConfig.ts
```diff
@@ -56,7 +60,8 @@ export default {
       "high-deal": "#E34A15",
       "fair-deal": "#EAA403",
       "good-deal": "#81B040",
-      "great-deal": "#149246",
+      "great-deal": colors.darkGreen,
+      "green-dark": colors.darkGreen,
```
lkappeler, line 64: Colour is reused but has a very specific name so I decide to add an export with a more generic name and keep backwards compatibility.

src/components/badges/gdbd.module.css
```diff
@@ -0,0 +1,19 @@
...
+.good-deal {
+  color: #81b040;
+}
+
+.fair-deal {
+  color: #eab000;
+}
+
+.high-deal {
+  color: #fa9000;
```
lkappeler, line 14: We do have tailwind values but unfortunately, they have been diverted for high-deal and not-defined and are consumed in other places of the project. Will create some tech task to clean up and define some common color palette

src/__tests__/__snapshots__/node.test.ts.snap
```diff
@@ -18,6 +18,7 @@ Object {
       "fair-deal": "#EAA403",
```
masone, line 18: I can't comment on the empty file above, so this one is out of place :D
care to delete the `Loading` file above? looks like an accident

src/components/badges/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Language } from "../../types"
+
+export type BadgeSize = "small" | "large"
+export interface BadgeProps {
+  size: BadgeSize
+  text: string
+  icon: ReactNode
+  tooltipContent?: ReactNode
+}
+export interface ExternalBadgeProps {
```
masone, line 17: why are the external? do you distinguish between internal/external ones from the pov of the package? 
maybe lend from our Foo/FooBase pattern that we often apply? Badge for the external BadgeBase for the shared one?

masone: I also feel that this type is misplaced. it has nothing to do with the component it is encapsulated with

src/tailwind/defaultConfig.ts
masone:
```suggestion
      "green-dark": colors.greenDark,
```

for the sake of consistency

masone: consider asking a designer what they call the color in the styleguide

src/types/index.tsx
```diff
@@ -0,0 +1 @@
+export type Language = "de" | "fr" | "it" | "en"
```
masone, line 1: very generic filename that will potentially attract a multiverse of different types. what about going for something more specific like `types/i18n.ts`?

(I realize just now that it's also a tsx file that is no component)

masone (approved):
I find that approach with the translations very straight forward and a fine compromise. This will help us harmonizing between DH/UL a lot. I imagine you won't be able to update DH as well as listings in one go - please make sure to create debug tasks for it (not Tech20) and plan them for team core in the next sprint

src/types/index.tsx
lkappeler: That was the idea to have a place to put shared types, no pref will rename and change the file

src/tailwind/defaultConfig.ts
lkappeler: Not available based on what I found on my own I think they reused the existing one made the same suggestion to Enrique to add it with the name `green-dark` to the additional colors section https://www.figma.com/file/Hg2Y3FaaStmRnswYAL9PY0/Core-UI-Components?node-id=0%3A1

src/__tests__/__snapshots__/node.test.ts.snap
lkappeler: Huch

src/assets/components/umbrella.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react"
+
+function UmbrellaIcon(props) {
```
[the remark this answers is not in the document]
lkappeler: 😁

src/components/badges/index.tsx
lkappeler: - Shared badge types no live in `badges/types.ts`
- Index has been renamed to base

src/tailwind/defaultConfig.ts
```diff
@@ -56,7 +60,8 @@ export default {
       "high-deal": "#E34A15",
       "fair-deal": "#EAA403",
       "good-deal": "#81B040",
-      "great-deal": "#149246",
+      "great-deal": colors.greenDark,
+      "green-dark": colors.greenDark,
```
masone, line 64: thinking about it... I think I'd prefer to stick to using the `great-deal` class instead of inventing our own names for colors that should be properly defined in the styleguide. I think it's wrong for the designer to use this color for this purpose in the first place

up to you.. :)

masone (approved):

src/tailwind/defaultConfig.ts
lkappeler: I checked it back with Enrique... (also synced quickly on the colours...)
image

PR #259, opened 2020-11-18
Update storybook

Marine-Berthier:
Reference  [CFYEO-13 ](https://autoricardo.atlassian.net/browse/CFYEO-13)

Major update on Storybook : new configuration + new story format
Upgrade from Storybook version 5.3.17 to Storybook version 6.0.28

:warning: This branch should not be merged until all the components have their new stories. 
The purpose of the PR is to adjust the setup before continuing. 

## TODO
- [ ] migrate all remaining components
- [ ] merge with semantic commit mesage


.storybook/addons.js
```diff
@@ -1,3 +1,3 @@
-import '@storybook/addon-actions/register';
-import '@storybook/addon-links/register';
-import '@storybook/addon-knobs/register';
+import '@storybook/addon-actions/register';
+import '@storybook/addon-links/register';
+import '@storybook/addon-knobs/register';
```
Averethel, line 3: Looks like your editor might be removing the newline at the end of the files, can you check the configuration?
The linter will complain about this.

.storybook/preview.js
```diff
@@ -0,0 +1,15 @@
...
+configure(require.context("../src", true, /\.stories\.mdx$/), module);
+
+global.___loader = {
+  enqueue: () => {},
+  hovering: () => {},
+};
+global.__PATH_PREFIX__ = "";
+window.___navigate = (pathname) => {
+  action("NavigateTo:")(pathname);
+};
```
masone, line 15: these all seem to be dummy implementations? do the docs suggest to do it this way or can we safely remove the noise?

package.json
```diff
@@ -71,6 +70,7 @@
     "react": "^16.12.0",
     "react-docgen-typescript-loader": "^3.6.0",
     "react-dom": "^16.12.0",
+    "react-is": "^17.0.1",
```
masone, line 73: I can't find this to be referenced in this PR. is it a necessary peer depencendy or an oversight?

package.json
```diff
@@ -43,18 +45,15 @@
...
     "@pika/plugin-copy-assets": "^0.9.0",
     "@pika/plugin-standard-pkg": "^0.9.0",
     "@pika/plugin-ts-standard-pkg": "^0.9.0",
-    "@storybook/addon-actions": "^5.3.17",
-    "@storybook/addon-info": "^5.3.17",
-    "@storybook/addon-knobs": "^5.3.17",
-    "@storybook/addon-links": "^5.3.17",
-    "@storybook/addons": "^5.3.17",
-    "@storybook/react": "^5.3.17",
-    "@storybook/storybook-deployer": "^2.8.6",
```
masone, line 57: removing the deployer might break some of the scripts the build pipeline relies on. 
we wanted to change the storybook deployment anyway, see https://autoricardo.atlassian.net/browse/CFYEO-12 - i'm ok with an intermediary solution when we tackle this next. happy to help with it

src/stories/Button.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+// also exported from '@storybook/react' if you can deal with breaking changes in 6.1
```
masone, line 2: can we deal with it and remove this? I don't understand what it wants to say :)

src/stories/Button.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
masone, line 1: regarding the filename - we have a `camelCase` convention for filenames. so this would be `stories/button.stories.tsx`. please check the other files too, just to make sure...

src/stories/page.css
```diff
@@ -0,0 +1,69 @@
+section {
+  font-family: 'Nunito Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
```
masone, line 2: Nunito Sans is not our brand font. is this stock-css that comes with storybook? I'd prefer to rely on our defaults from https://github.com/carforyou/carforyou-components-pkg/blob/master/assets/global.css, which is also what we're using in the projects, so default fonts and sizes would be consistent

src/stories/button.css
```diff
@@ -0,0 +1,30 @@
...
+  padding: 10px 16px;
+}
+.storybook-button--medium {
+  font-size: 14px;
+  padding: 11px 20px;
+}
+.storybook-button--large {
+  font-size: 16px;
+  padding: 12px 24px;
+}
```
masone, line 30: is it possible to omit all of these style files for styling storybook itself and rely on some kind of internal defaults?

src/stories/Page.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
```
masone, line 1: to be cleaned up :)

masone (commented):
looks generally a-ok. there are a lot of changes I can't put in perspective. how would you suggest to proceed?

.storybook/addons.js
Marine-Berthier: Yes, I changed the configuration, it should be fixed.

.storybook/preview.js
Marine-Berthier: You're right ! I removed it and more.

package.json
Marine-Berthier: How can I check it this is a necessary peer dependency ? Can I be sure by checking the "requires" sections in package-lock ?

src/stories/Button.stories.tsx
Marine-Berthier: This files and the one below were actually examples from Storybook. I forgot to remove them. Sorry about that :confused:

package.json
Marine-Berthier: I re-installed the deployer package (before finding another solution)

.storybook/main.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  stories: ['../src/**/*.stories.@(js|mdx)'],
```
masone, line 2: do we need `tsx` here in addition?

.storybook/preview.js
```diff
@@ -0,0 +1,6 @@
+import { action } from "@storybook/addon-actions";
+import { configure } from "@storybook/react";
+
+import "./stories.css";
+
+configure(require.context("../src", true, /\.stories\.mdx$/), module);
```
masone, line 6: this feels like a duplication of `stories: ['../src/**/*.stories.@(js|mdx)'], ` in main.js - do we need both? what is each for?

package.json
masone: you can delete the line and run `npm install` - this would warn you about missing peer dependencies if it's required by any package that we're using. if you didn't add it intentionally, it might very well be that it slipped in somehow

src/stories/assets/stackalt.svg
```diff
@@ -0,0 +1 @@
...1 L37.0505103,32.4489717 C37.6150232,34.5847667 36.3412474,36.7737988 34.2054524,37.3383117 C33.5569474,37.5097184 32.87575,37.515287 32.2245294,37.3545053 L23.3459013,35.1624353 C18.2976962,33.9160695 14.3561301,29.9745034 13.1097642,24.9262983 L10.9176943,16.0476701 C10.3881737,13.9029317 11.6975664,11.7350164 13.8423049,11.2054957 C14.4935255,11.044714 15.1747229,11.0502826 15.8232279,11.2216893 Z" opacity=".5" transform="translate(23.999997, 24.272222) rotate(-45.000000) translate(-23.999997, -24.272222)"/><path id="Rectangle-46-Copy" fill="#FC521F" d="M15.8232279,3.32785281 L24.7368455,5.68381509 C29.6053842,6.97062075 33.4077423,10.7729788 34.694548,15.6415176 L37.0505103,24.5551352 C37.6150232,26.6909302 36.3412474,28.8799623 34.2054524,29.4444752 C33.5569474,29.6158819 32.87575,29.6214505 32.2245294,29.4606688 L23.3459013,27.2685988 C18.2976962,26.022233 14.3561301,22.0806669 13.1097642,17.0324618 L10.9176943,8.15383364 C10.3881737,6.00909519 11.6975664,3.84117987 13.8423049,3.31165925 C14.4935255,3.15087753 15.1747229,3.15644615 15.8232279,3.32785281 Z" opacity=".5" transform="translate(23.999997, 16.378385) rotate(-45.000000) translate(-23.999997, -16.378385)"/></g></svg>
```
masone, line 1: I would vote do delete everything we don't need. some of these look like storybook example icons?

tsconfig.json
```diff
@@ -4,10 +4,10 @@
     "target": "es5",
     "lib": ["es5", "es6", "es7", "es2017", "dom"],
     "jsx": "react",
-    "rootDirs": ["src", "stories"],
+    "rootDirs": ["src", "src/stories"],
     "baseUrl": "src",
     "esModuleInterop": true
   },
-  "include": ["typings.d.ts", "src/**/*", "stories/**/*"],
+  "include": ["typings.d.ts", "src/**/*", "src/stories/**/*", "src/stories"],
```
masone, line 11:
```suggestion
  "include": ["typings.d.ts", "src/**/*"],
```

should match both `src/stories` and `src/stories/**`. can you try to simplify and see if it still works?

tsconfig.json
```diff
@@ -4,10 +4,10 @@
     "target": "es5",
     "lib": ["es5", "es6", "es7", "es2017", "dom"],
     "jsx": "react",
-    "rootDirs": ["src", "stories"],
+    "rootDirs": ["src", "src/stories"],
```
masone, line 7:
```suggestion
    "rootDirs": ["src"], 
```

should be sufficient. before the folders were on the same level, now stories are a subpath of the already included directory

masone (commented):
thanks for cleaning up. definitely helps with understanding what's going on :)
I would recommend to get it into a "mergeable shape" from here on, ie. get all PR checks to pass. then, let's take a step back and discuss how to proceed

package.json
Marine-Berthier: I checked, it is a peer dependency for addon-docs and addon-essential

.storybook/main.js
Marine-Berthier: We might, I added it.

.storybook/preview.js
Marine-Berthier: You are right, this line is not needed. 
In the previous version we only had 1 file called config.js to configure everything, now we have main.js to indicate the location of the story files and the list of addons used. Preview.js configures the iframe that renders components, it is the file where we need to import global stylesheet for exemple.

src/stories/button.css
ralucacril: i think we should refactor this soon due to our naming conventions. the BEM... should be `storybook-button_large`
but is not part of this ticket though 😢

Marine-Berthier:
@masone, @ralucacril : I removed knobs. Label is now part of the arguments and the code preview includes it as a props (the same for buttonName). Should I try to find a better option ?
SB

Averethel:
I'm not following the progress here and don't know what are the best practices in v6, so I might be off here.
In general, knobs are great if someone wants to play around and see how the component looks with different props. We tend to have stories for every major variant but I'm not sure if we want to cover every possible combination.
The way I see it the main users of knobs would be:
- engineers who are going to use a more complex component for the first time and want to play around with it before implementing
- designers when checking which components are available and verifying the implementation with the design system
We might not need it short term but it would be beneficial to have.

Marine-Berthier:
In v6 Storybook introduced the usage of "args" (= component's props). For each story the state of the component is described using these args and the user can interact with them dynamically. It will be easier to play with components.
From what I understood everything can be added as an argument (within *.stories.js) but if someone wants to copy/paste the code then there will be additional arguments which are not part of the component's props. Maybe no one uses it like this and then it is not an issue but I wanted to point this :smile:

package.json
```diff
@@ -68,9 +68,10 @@
     "marked-terminal": "^4.1.0",
     "postcss-easy-import": "^3.0.0",
     "postcss-loader": "^3.0.0",
-    "react": "^16.12.0",
+    "react": "^16.14.0",
     "react-docgen-typescript-loader": "^3.6.0",
     "react-dom": "^16.12.0",
+    "react-is": "^16.13.1",
```
masone, line 74: what is this used for?

src/stories/button.stories.js
```diff
@@ -0,0 +1,71 @@
...
+        disable: true,
+      },
+    },
+  },
+}
+
+const Template = (args) => {
+  return (
+    <div className="mx-30 mb-40">
+      <div className="text-2xl mb-20">{args.buttonName}</div>
```
masone, line 24: we could extract the title and the container into some helper component that we use for displaying our stories. I found myself repeating this when migrating another component

src/stories/button.stories.js
```diff
@@ -0,0 +1,71 @@
+import React from "react"
+import Button from "../components/button"
+import { action } from "@storybook/addon-actions"
+
+export default {
+  title: "Button",
+  component: Button,
+  args: {
+    buttonName: "Default",
```
masone, line 9: this is related to the comment below.. if we decide to add the title by convention, we could generalize the name here a bit, like`componentTitle`

masone (commented):
I'm really happy with the current result. Let's continue migrating stories and see what pops up...

package.json
```diff
@@ -30,7 +30,9 @@
     "tailwindcss": "^1.4.0"
   },
   "dependencies": {
+    "bindings": "^1.5.0",
```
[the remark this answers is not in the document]
Marine-Berthier: Yesterday I had an error when uninstalling bindings and react-is, today it seems to work fine without them :confused:

src/stories/button.stories.js
Marine-Berthier: I tried something but it was not conclusive. I am not sure how to this

masone: let's migrate 2-3 components and a solution might present itself. I have something in mind that I could make a proposal for

package.json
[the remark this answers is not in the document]
masone: 🤷‍♂️ :)

the only thing I can imagine is that it's a peer dependency of something. it would fail when you ran `npm ls` and tell you to install it

src/stories/iconButton.stories.js
```diff
@@ -0,0 +1,42 @@
...
+  label: "Anrufen",
+}
+
+export const WrappingALink = Template.bind({})
+WrappingALink.args = {
+  icon: () => <CtaCall />,
+  onClick: action("on Click"),
+  teal: false,
+  storyName: "Wrapping a link",
+  label: <a href="https://carforyou.ch">CarForYou</a>,
```
masone, line 41: search for all occurences :)

src/stories/intercom.stories.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+const Template = (args) => {
+  return (
+    <div className="mx-30 mb-40">
+      <div className="text-2xl mb-20">{args.storyName}</div>
+      <div className="w-12/12 md:w-3/12">
+        <Intercom
+          {...args}
+          label={("label", "Support")}
```
masone, line 26: this looks weird, can you check what this is supposed to do? :)

src/stories/intercom.stories.js
```diff
@@ -0,0 +1,43 @@
...
+          appId={
+            ("appId (you need am actual app id to be able to load intercom)",
+            "123")
+          }
+          stage={("stage", "dev")}
+          autoload={("autoload", false)}
```
masone, line 32: here as well, maybe do some search `{("` and `")}`

masone (commented):
as a separate task, let's migrate all stories to typescript

package.json
```diff
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.6",
+    "core-js": "^3.7.0",
```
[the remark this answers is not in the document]
Marine-Berthier: I wanted to double check to which dependency it was a peer dependency : it seems that it is none. I don't have any errors or warning on `npm install` so I uninstalled it

src/stories/alertMessage.stories.js
```diff
@@ -0,0 +1,94 @@
...
+    fullWidth: false,
+    alignCenter: false,
+  },
+}
+
+const Template = (args) => {
+  return (
+    <div className="mx-30 mb-40">
+      <div className="text-2xl mb-20">{args.alertName}</div>
+      <div className="w-12/12">
```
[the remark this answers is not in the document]
Marine-Berthier: I stored it in a file called generator to avoid repeating those lines

src/stories/intercom.stories.js
Marine-Berthier: To me it is just regular arguments so I rewrote it the "normal" way

src/stories/spinner.stories.js
```diff
@@ -0,0 +1,24 @@
...
+    storyName: {
+      table: {
+        disable: true,
+      },
+    },
+  },
+}
+
+const Template = (args) => {
+  return <Generator title={args.storyName} style={""} component={<Spinner />} />
```
masone, line 21:
```suggestion
  return <Generator title={args.storyName} component={<Spinner />} />
```

src/stories/spinner.stories.js
```diff
@@ -0,0 +1,24 @@
+import React from "react"
+import Generator from "./generator.js"
```
masone, line 2: we should be able to drop the `.js` extension from the import. really not a biggie - just for the sake of consistency with the other imports
```suggestion
import Generator from "./generator"
```

src/stories/intercom.stories.js
```diff
@@ -0,0 +1,33 @@
...
+    buttonName: {
+      table: {
+        disable: true,
+      },
+    },
+  },
+}
+
+const Template = (args) => {
+  return <Generator title={args.storyName} component={<Intercom {...args} />} />
```
masone, line 26: generator is quite a generic name. maybe something like...
```suggestion
  return <StoryContainer title={args.storyName} component={<Intercom {...args} />} />
```

masone (approved):
all remaining input is purely style or personal preference. please move on with merging as soon as you're happy with the result. because it's such a large change, I'd get someone elses approval in addition :)

super nice job! thanks for taking care of this tedious update. I'm happy you invested the time as the result is so much cleaner than it was before

ralucacril (approved):
amazing effort! 👏  thanks a lot for driving this and make it happen! 🎉

PR #267, opened 2020-11-25
Storybook update : dropdown component

Marine-Berthier:
Stories for the dropdown component written in v6 format

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "tailwindcss": "^1.4.0"
   },
   "dependencies": {
-    "bindings": "^1.5.0",
     "classnames": "^2.2.6",
     "core-js": "^3.7.0",
     "deepmerge": "^4.2.2",
@@ -71,7 +70,6 @@
     "react": "^16.14.0",
     "react-docgen-typescript-loader": "^3.6.0",
     "react-dom": "^16.12.0",
-    "react-is": "^16.13.1",
     "react-test-renderer": "^16.12.0",
     "semantic-release": "^17.0.0",
     "svgo": "^1.3.2",
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "tailwindcss": "^1.4.0"
   },
   "dependencies": {
+    "@storybook/addon-knobs": "^6.1.6",
     "classnames": "^2.2.6",
     "core-js": "^3.7.0",
     "deepmerge": "^4.2.2",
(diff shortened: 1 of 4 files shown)


ralucacril:
you might need to merge the original branch back into yours since you have some conflicts. nice work! 🎉  👏

ralucacril (approved):

package.json
```diff
@@ -30,7 +30,7 @@
     "tailwindcss": "^1.4.0"
   },
   "dependencies": {
-    "bindings": "^1.5.0",
+    "@storybook/addon-knobs": "^6.1.6",
```
masone, line 33: seems like you were not on the latest version when you branched off the branch `update-storybook`? you can try 
```
git checkout update-storybook
git pull
git checkout dropdown
git rebase update-storybook
```

Marine-Berthier:
I merged the original branch to `dropdown`, it seems there's no more conflict. Should I merge right now ?

src/stories/dropdown.stories.js
```diff
@@ -0,0 +1,116 @@
+import React from "react"
+import Dropdown from "../components/dropdown/index"
+import { action } from "@storybook/addon-actions"
+import { object, number, boolean, text } from "@storybook/addon-knobs"
```
masone, line 4: I understood we wanted to go without knobs? did you find them to be valuable after all and we actually want them in?

ralucacril: i think @Averethel said he sees values in knobs

Marine-Berthier: For this particular component I could not find another way to make it work, but I can try harder :blush: However there is no Knobs pannel created, everything is under Controls pannel.

Marine-Berthier: Actually, everything works perfectly fine without them 🤦 I was looking for something complicated when actually it was very simple ! Should I push the changes (stories without knobs) ?

Averethel: @ralucacril I did, but @Marine-Berthier explained that there's now a different way to play around with component. Let's roll without it for now and if we see we need them - we act.

ralucacril: a ok. my bad. then @Marine-Berthier roll them without the knobs if is working

Marine-Berthier: I realised with this component that I didn't really understand the purpose of knobs. I thought that knobs could be replaced by args as soon as it is now possible to play with args the same way you can play with knobs. Actually it is still 2 different things : knobs could still be used and won't appear as args in the code. I might be wrong though

PR #269, opened 2020-11-25
Migrate icon button

ralucacril:


src/stories/iconButton.stories.js
```diff
@@ -0,0 +1,42 @@
...
+import { action } from "@storybook/addon-actions"
+
+import Button from "../components/button"
+import CtaCall from "../../.storybook/icons/ctaCall"
+
+export default {
+  title: "Icon Button",
+  component: Button,
+  args: {
+    storyName: "Default",
```
Marine-Berthier, line 11: There's no need to specify `label` here ? I did it for the button component but it might not be the best way actually.

src/stories/iconButton.stories.js
```diff
@@ -0,0 +1,42 @@
...
+  )
+}
+
+export const Default = Template.bind({})
+Default.args = {
+  icon: () => <CtaCall />,
+  onClick: action("on Click"),
+  teal: false,
+  storyName: "Default",
+  label: "Anrufen",
```
ralucacril, line 32: i added the label here. I didn't fully udnerstand what the label from above does..

ralucacril:
and i merged by mistake...i thought you approved. 😢

PR #273, opened 2020-11-26
fix: whitelist scrolling classes

Averethel:
Fixes CAR-5561

The `preventScrolling` classes are dynamic and therefore stripped by purgeCSS.


ralucacril (approved):
ups...glad that you find this

masone:
oopsies! 🙊

PR #283, opened 2020-11-30
Migrate alert message

masone:


src/stories/alertMessage.stories.js
```diff
@@ -0,0 +1,61 @@
...
+  alignCenter: true,
+  fullWidth: true,
+}
+
+export const FullWidth = Template.bind({})
+FullWidth.args = {
+  alertName: "Full width",
+  alignCenter: false,
+  fullWidth: true,
+}
```
Marine-Berthier, line 61: It looks like some stories are missing : `success` and the ones `with icons`. Regarding the `type`, do you think a dropdown within the control pannel could be enough ? Something like the one to select `alignment` and` position` for the [`Tooltip`](https://github.com/carforyou/carforyou-components-pkg/pull/286/files)

Marine-Berthier:
@masone I worked directly on your branch and so pushed my changes directly here : was it the right/best way to do it ?

masone:
I can't approve my own PR, so you'll have to do it :)
Then, feel free to merge!

Marine-Berthier (approved):

masone:
thanks for the support Marine

PR #284, opened 2020-12-01
Migrate select

Marine-Berthier:


src/stories/select/simple/appearance.stories.js
```diff
@@ -0,0 +1,88 @@
...
+      { value: 2, name: "Two" },
+      { value: 3, name: "Three" },
+      { value: 4, name: "Four" },
+      { value: 5, name: "Five" },
+      { value: 6, name: "Six" },
+    ],
+    allowCustomValues: false,
+    skipContainer: false,
+    withAutosuggest: false,
+    handleChange: handleChange(),
```
ufkaya, line 23: here you are calling the fuction `handleChange`, is it correct? Or you want just to assign it?

masone (approved):
thanks for explaining what we need the state for in https://autoricardo-ag.slack.com/archives/C01CL9JTTUP/p1607445083004400

let's roll with it then!

PR #285, opened 2020-12-02
Migrate checkbox

Marine-Berthier:


.storybook/manager.js
```diff
@@ -0,0 +1,2 @@
+import { addons } from "@storybook/addons";
+addons.setConfig({ showRoots: false });
```
masone, line 2: I'm surprised the linter does not yell at us for using semicolons at the end of the line. it's something we can look at in a separate PR

.storybook/preview.js
```diff
@@ -1,2 +1,4 @@
-import "./stories.css"
+import "./stories.css";
```
masone, line 1: same as above. the style linter should complain here. I'll create a ticket for it

src/stories/checkbox/checked.stories.js
```diff
@@ -0,0 +1,96 @@
...
+  const [value, setValue] = useState(true)
+  return (
+    <div className="mx-30 mb-40">
+      <div className="text-2xl mb-20">Example</div>
+      <div className="w-12/12 md:w-3/12">
+        <Checkbox
+          {...args}
+          checked={value}
+          onChange={(e) => {
+            setValue(e.target.checked)
```
masone, line 34: imo, this state should be internal to the `Checkbox`. there should be no need to keep state externally and I would certainly not encourage it in example code. can you check if this is just a port of a prior implementation or required in your opinion?

src/stories/checkbox/unchecked.stories.js
```diff
@@ -0,0 +1,96 @@
+import React, { useState } from "react"
```
masone, line 1: I don't think there's much value in duplicating all stories just for the sake of showing them in checked / unchecked state. one can do that by just checking the checkbox in the other file... also, there are variations for checked/unchecked already above

what do you think about dropping these? maybe you see value I'm missing....

src/stories/checkbox/checked.stories.js
```diff
@@ -0,0 +1,96 @@
...
+  storyName: "Button style disabled",
+}
+
+export const ButtonStyleWithErrorMessage = Template.bind({})
+ButtonStyleWithErrorMessage.args = {
+  buttonStyle: true,
+  label: "Label",
+  error: "Error message",
+  storyName: "Button style with error message",
+}
```
masone, line 96: as these have a complete different visual representation, it might be worth putting them in a separate story

src/stories/checkbox/checked.stories.js
Marine-Berthier: Do you mean that useState should not be used within the stories ?

src/stories/checkbox/unchecked.stories.js
Marine-Berthier: I think you are right.  As it is right now the box is checked and unchecked on click and the` checked `args in the control pannel does not change accordingly. Will try to fix that.

Marine-Berthier:
There's now 2 files : one for the default checkbox and one for the button style checkbox. 
No more state involved but it is only possible to check and uncheck the box in the control pannel, click does not work. Do you think there's a way to make both change options work ? I could not find anything but will keep on looking :mag_right:

masone (approved):

PR #286, opened 2020-12-03
Tooltip

Marine-Berthier:
For this component I created only one story since it is now possible to play around with the tooltip's position and alignment in the Control pannel. If you think we still need to write a story for every position and alignment let me know

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,1 +1,2 @@
-import "./stories.css";
+import "./stories.css"
+import "../assets/index.css"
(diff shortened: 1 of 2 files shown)


ralucacril:
I think you are missing some stories.
Screenshot 2020-12-03 at 17 08 10

Marine-Berthier:
@ralucacril yes, that's what I explained in the comment above. I am not sure but since it is now possible to play around with the tooltip's position and alignment in the Control pannel, do we still need the whole list ?

ralucacril:
so so sorry, i missed your comment. 🙈

ralucacril:
tbh for us might not be important to have them as a separate story. but maybe for designers could be important. might not hurt to ask Fred (he is on slack)  or Enrique if they would find the benefit of the story.

Marine-Berthier:
@ralucacril Thanks Raluca, I will ask Enrique 👍

Marine-Berthier:
@ralucacril According to Enrique there's no need to have the full list

ralucacril (approved):
thanks for checking with Enrique

PR #288, opened 2020-12-07
Migrate radiobutton

Marine-Berthier:
Stories for both `selected` and `notSelected` radio button, however only one file could be enough as it is possible to set `checked` to true or false within the control pannel.


.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
masone, line 2: this file contains configuration for your local development environment. not everyone will want the same settings, so I'd gitignore this file. because the file is not project-specific, but rather specific to the setup of the developer, we tend to not put these into the projects `.gitignore`, but a global one on your local machine. see how to set one up here: https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#create-a-global-gitignore

as the file is empty, you can as well just delete it :)

src/stories/radioButton/notSelected.stories.js
```diff
@@ -0,0 +1,71 @@
...
+const onChange = () => action("onChange")
+const initialValue = "B"
+
+export default {
+  title: "Radio Button/Not selected",
+  component: RadioButton,
+  args: {
+    storyName: "",
+    value: initialValue,
+    checked: false,
```
masone, line 14: we discussed this in another PR, but maybe I missed the decision. I really don't see the value of adding all variations of checkboxes with checked/unchecked. a knob for checking each would be enough for me

masone: Or have one checked representation of the default one. If I remember correctly, the checked checkboxes have all the same style in each variation

.vscode/settings.json
Marine-Berthier: Thanks for the clarification @masone 👍

src/stories/radioButton/notSelected.stories.js
Marine-Berthier: We had this discussion for the Checkbox and decided to not show each variation. 
I thought of doing the same for the RadioButton but I doubted as it behaves differently when checked or unchecked : the click on an unchecked radioButton will trigger an action but the click on a checked one won't trigger anything.
Actions might not be that important in Storybook though. What do you think ?

masone: Yeah, I would do the same for radiobutton and not do checked examples

PR #293, opened 2020-12-15
CFYEO-26: extract collapsible

talamcol:
References [CFYEO-26](https://autoricardo.atlassian.net/browse/CFYEO-26)

The listings project and the dealer hub shared some common logic to collapse/expand elements. This PR extracted the component to the component-pkg.

- Moved the collapsible component from the listing project to this project and refactored some parts of the code
- Moved the test for the collapsible component
- Created a new storybook page


src/components/__tests__/collapsible.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+      act(() => {
+        fireEvent.click(getByText(openedText))
+      })
+
+      expect(getByText(closedText))
+      expect(queryAllByText(openedText)).toEqual([])
+      expect(queryAllByText(content)).toEqual([])
+    })
+  })
+})
```
masone, line 70: nice, thanks for adding tests!

src/components/collapsible.tsx
```diff
@@ -0,0 +1,74 @@
...
+  renderToggle: (options: { isCollapsed: boolean }) => ReactNode
+  /**
+   * Children to be rendered within the collapse component. The elements will be visible when the component is expanded.
+   */
+  children?: () => ReactNode
+  /**
+   * Function executed when the collapse state has changed.
+   *  - passes the new collapse state as a function parameter
+   */
+  onCollapseStateChange?: (isCollapsed: boolean) => void
```
masone, line 25: For the sake of brevity, we could also just call it `onChange`. in the context of `Collapsible`, this seems like the only state that can ever change
```suggestion
  onChange?: (isCollapsed: boolean) => void
```

masone: In `renderToggle`, we're passing an options object with `isCollapsed` to the callback. shall we keep the two consistent? either always pass a variable or always pass an object?

src/components/collapsible.tsx
```diff
@@ -0,0 +1,74 @@
...
+   */
+  opacityOnHover?: boolean
+}
+
+const Collapsible: FC<Props> = ({
+  renderToggle,
+  isInitiallyCollapsed = true,
+  children,
+  onCollapseStateChange,
+  collapseIconSize = 32,
```
masone, line 43: I dislike passing exact sizes in pixels as props. The reason is that it's pretty easy to misuse it and allow all sorts of icon sizes, while the styleguide actually limits them. Especially in a shared component package, I would recommend to be as strict as possible and not allow any unwanted values

How many sizes are you experiencing in the projects? Can we settle for a toggle like `"large" | "small"` or something?

src/stories/collapsible.stories.js
```diff
@@ -0,0 +1,50 @@
...
+
+export default {
+  title: "Collapsible",
+  component: Collapsible,
+  args: {
+    isInitiallyCollapsed: true,
+    renderToggle: (options) => options,
+    onCollapseStateChange: (isCollapsed) => isCollapsed,
+    collapseIconSize: 32,
+    opacityOnHover: true,
```
masone, line 14: this is part of our default link style. I would quickly align with the designer. we might want either or, but not switch it on/off per case. this is a great moment in time to align the hover style of all collapsibles instead of allowing both variants

src/stories/collapsible.stories.js
```diff
@@ -0,0 +1,50 @@
...
+      Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy
+      eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam
+      voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
+      clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit
+      amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam
+      nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed
+      diam voluptua
+    </div>
+  ),
+}
```
masone, line 50: sweet, thanks for adding the stories!

masone (changes requested):
Looks really thorough and nice, thanks! 
I have a couple of things to discuss before we proceed. For extracted components, we try to set the standards high. As they'll be used across projects it's worth spending some extra time on making sure we're enforcing consistency :)

src/components/collapsible.tsx
ralucacril: to be honest i would argue why is neccesary to have different sizes if the icons. what are the use cases for having different sizes? afaik after discussing with designers is that if we have an icon of 24px we should not make it bigger to like 32px, but we should use a different icon due to the stroke.

talamcol: we only use 24, 32 and 40 (dealerhub listing overview page). So I could go with "small", "medium", and "large".

talamcol: > to be honest i would argue why is neccesary to have different sizes if the icons. what are the use cases for having different sizes? afaik after discussing with designers is that if we have an icon of 24px we should not make it bigger to like 32px, but we should use a different icon due to the stroke.

because dealerhub uses a larger icon on the overview page. That's why I am having this as a prop. 
On listings, the icon is always 32x32 as it is at the moment

ralucacril: i don't think we should use 40px at all. i think we should check with the designer the cases we use 24, 32 and 40 and make the uniform. 40 is definitely not a number to use.
and regarding using 24 and 32, as i mentioned before, they suggest using 2 different icons due to the stroke. 

my advice would be to check with Fred or Enrique the cases where we use 32 and 40 and try to see if we can use 24 in both case.

masone: fair. I would still bring it to the attention of a designer, for example Enrique and verify their intent was to have three sizes. often they're not aware and very happy to simplify the design system

src/stories/collapsible.stories.js
talamcol: this is also only because of the dealerhub overview page. I discussed this one with @lkappeler . The `a` tag has a default hover styling with `opacity: 60`. it looks a bit strange if the whole component gets transparent on the listing overview

src/components/collapsible.tsx
talamcol: good point. will change it

src/components/collapsible.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { ReactNode, FC, useState } from "react"
+import classNames from "classnames"
+
+import ArrowDown from "./icons/arrowDown"
+
+interface Props {
+  /**
+   * False if the element shall be expanded by default
+   * default: true
```
Averethel, line 9: Unless this changed in storybook 6 you don't need to provide the default here. It will be taken from the default values assignment in the component function.

src/components/collapsible.tsx
```diff
@@ -0,0 +1,74 @@
...
+   *  - isCollapsed tells you the current state of the collapsible element
+   */
+  renderToggle: (options: { isCollapsed: boolean }) => ReactNode
+  /**
+   * Children to be rendered within the collapse component. The elements will be visible when the component is expanded.
+   */
+  children?: () => ReactNode
+  /**
+   * Function executed when the collapse state has changed.
+   *  - passes the new collapse state as a function parameter
```
Averethel, line 23:
```suggestion
   * Function executed when the collapsible state has changed.
   *  - passes the new collapsible state as a function parameter
```

src/components/collapsible.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [isCollapsed, setIsCollapsed] = useState(isInitiallyCollapsed)
+
+  return (
+    <>
+      <a
+        className={classNames("flex items-center cursor-pointer", {
+          "hover:opacity-100": !opacityOnHover,
+        })}
+        onClick={() => {
+          onCollapseStateChange ? onCollapseStateChange(!isCollapsed) : null
```
Averethel, line 55: I'd go for:
```suggestion
          onCollapseStateChange && onCollapseStateChange(!isCollapsed)
```

src/components/icons/arrowDown.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react"
```
Averethel, line 1: Hmmm do we not build components from automatically here yet?

src/components/collapsible.tsx
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Function executed when the collapse state has changed.
+   *  - passes the new collapse state as a function parameter
+   */
+  onCollapseStateChange?: (isCollapsed: boolean) => void
+  /**
+   * The size of the arrow down icon
+   * default: 32
+   */
+  collapseIconSize?: number
```
Averethel, line 30: Any chance we can go with something like `small` | `large` instead of pixel size?

src/components/collapsible.tsx
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * If the element should get a bit transparent on hover
+   * default: true
+   */
+  opacityOnHover?: boolean
```
Averethel, line 35: Shouldn't this always be opaque on hover?

Averethel (commented):
Looks pretty solid. Good job with tests and stories. I have a couple of suggestions:

src/components/icons/arrowDown.tsx
talamcol: no, not yet

src/components/collapsible.tsx
talamcol: @Averethel  what do you think if I just restrict it to the allowed numbers?

`collapseIconSize?: 24 | 32 | 40`

src/components/collapsible.tsx
talamcol: I use the component also in the listing overview in dealerhub. It looks a bit strange if it has an opacity:
image

src/components/collapsible.tsx
talamcol: image

As we currently only use 32px in both projects, I'll drop the prop completely

talamcol: I also imported the new icon he sent me from here: https://www.figma.com/file/L06vFtqz3AhJbbdiGWPuzy/Myicons-v1.16?node-id=40%3A1986 to be completely up to date 👍

ralucacril: thanks a lot for aligning this 🙏

src/components/collapsible.tsx
talamcol: removed it completely after discussion with UX

src/components/icons/arrowDown.tsx
Averethel: Can you create a follow up for this? This is something that we use in other projects, makes sense to align the approach

talamcol: Created a task for it: https://autoricardo.atlassian.net/browse/CAR-5906

src/components/__tests__/collapsible.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+  })
+
+  describe("on change function", () => {
+    it("should call the onChange function if defined", () => {
+      const customOnChangeFunction = jest.fn()
+      const { getByText } = renderWrapper({
+        onChange: customOnChangeFunction,
+      })
+
+      act(() => {
```
masone, line 77: I think to remember that most of the react-testing-library helpers are wrapped with `act` internally and most of the time, we don't need to `act` in the test ourselves. can you confirm why we need to to do explicitly here?

src/stories/collapsible.stories.js
masone: I'm sure there's a way to disable the default opacity on the parent DOM element instead of leaking this project-specific weirdness into the components-pkg. Can we try to get rid of this extra prop?

talamcol: unfortunately, CSS does not support parent selectors, only sibling and child. I can try to use a `ref` and then set the style dependent if the ref on the child is set or not. But this probably makes the code more complex than it is now

src/components/__tests__/collapsible.test.tsx
talamcol: no we don't. This was before my deep dive into the testing library during the holidays. `.click()` already wraps everything into act. Removed it

src/stories/collapsible.stories.js
masone: I meant... if there's a default opacity, we might be able to override it on the parent. the most naive approach in pseudocode:

```
opacity-0 !important (add a new wrapper that overrides the behavior of the child)
  opacity-60 (I assume this is what's in the children)
```

as this thread starts to get long, it might be a good time to quickly align in a call. just ping me!

masone: We just discussed this on the phone. Opacity is relative to the parent, so we can't override it on the child. We get the feeling that the current approach is the most descriptive we can come up with. I'm fine moving on, despite that the prop feels a bit out-of-place and leaks project-specific global css into the component. Let's roll with this if we see no other way out

masone (approved):

PR #294, opened 2020-12-21
feat: new custom isMounted hook

talamcol:
References: [CFYEO-37](https://autoricardo.atlassian.net/browse/CFYEO-37)

**Motivation**
If an async function is called within a useEffect hook and the value will be used to update a state, it may happen that the component is already unmounted when the data is received. -> please check https://github.com/carforyou/carforyou-dealerhub-web/pull/880/files#r551746981 to see how it can be used.

This could yield to memory leaks in the application. Therefore, I created a new hook to ensure that. It will be used in dealerhub and listings to fix the unit test warnings

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,6 +15,7 @@ import RadioButton from "./components/radioButton"
 import WithLabel from "./components/fieldHelpers/withLabel"
 import WithValidationError from "./components/fieldHelpers/withValidationError"
 import useModal from "./hooks/useModal"
+import useIsMountedRef from "./hooks/useIsMountedRef"
 import Profile from "./components/icons/profile"
 import LogoRedWhite from "./components/icons/logoRedWhite"
 import LogoRedBlack from "./components/icons/logoRedBlack"
@@ -44,6 +45,7 @@ export {
   WithLabel,
   WithValidationError,
   useModal,
+  useIsMountedRef,
   Profile,
   LogoRedWhite,
   LogoRedBlack,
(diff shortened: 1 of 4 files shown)


lkappeler (commented):
The code looks good to me. I am not sure weather the components library is the right place for the custom hooks as it does not really provide a UI component. We do have similar issues when it comes to sharing library code between projects as we do not really have a place for them...
Maybe its a good time to bring the discussion up in the team and agree on how to share "utility" code between projects...

src/index.ts
```diff
@@ -44,6 +45,7 @@ export {
   WithLabel,
   WithValidationError,
   useModal,
+  useIsMountedRef,
```
ralucacril, line 48: maybe i dont understand, but where do we use it?

ralucacril: aa nevermind. is the components library. 🤦  but i do agree with @lkappeler i think we should reconsider the place for this. maybe we can have a discussion about this in an RFC?

masone: it was my first reaction as well, but in the end we're sharing hooks already now. they're only gonna be used in components, so this feels like the right place to me. I would not want an extra package just for hooks and then end up consuming both packages in most projects anyway :)

(just my 2 cents. I'm open to exploring alternative ways)

masone (commented):
Can you give a bit more context please? The description is slightly confusing to me. How would one do state updates when the component is not mounted?

talamcol:
> Can you give a bit more context please? The description is slightly confusing to me. How would one do state updates when the component is not mounted?

@masone 
Updated the description and referenced a related comment in another PR.

src/index.ts
talamcol: @ralucacril I used it here https://github.com/carforyou/carforyou-dealerhub-web/pull/880

The location seems odd but there is already the `useModal` hook inside the component package. I'm fine with removing it but eventually, we will use it in listings and dealerhub

talamcol:
Closing this PR as per discussion in https://github.com/carforyou/carforyou-dealerhub-web/pull/880#discussion_r551772570

PR #297, opened 2021-01-08
CFYEO-39: Remove classname props

talamcol:
Reference: [CFYEO-39](https://autoricardo.atlassian.net/browse/CFYEO-39)

Some components provided the ability to pass classes as a prop. This gives the consumer of that component a lot of possibilities to mess around with the styling. Therefore, I removed them where possible.

Unfortunately, I was not able to remove the menuClassName prop as it will require a lot of restructuring or inventing new boolean props to achieve the same styling. However, I think this change provides a bit of value nevertheless as it improves consistency and does not allow className in other places now.

----

Here are the related PR's:
- dealerhub: https://github.com/carforyou/carforyou-dealerhub-web/pull/894
- listings: https://github.com/carforyou/carforyou-listings-web/pull/2801
- header-footer: not affected by this change
- content-web: not affected by this change
- email-templates: not affected by this change
 
----

The following was discussed and approved with the UX designer:

1. Shadow hard on full-screen dropdowns -> aligned with the dropdown of autosuggest fields
Before:
image

Now:
image

2. Added border-radius to full-screen dropdowns
Before:
image

Now:
image

(you have to look very closely and it's not that obvious)

### DIFF
--- a/assets/input/basic.css
+++ b/assets/input/basic.css
@@ -88,22 +88,12 @@ textarea[data-valid="false"]:focus {
 }
 
 input.input_left {
-  @apply rounded-l-4 rounded-r-none;
+  @apply rounded-r-none;
 }
 
 input.input_right,
 div.input_right {
-  @apply rounded-r-4 rounded-l-none -ml-1;
-}
-
-.input_left:hover,
-.input_right:hover {
-  @apply z-1 transition duration-200;
-}
-
-.input_left:focus,
-.input_right:focus {
-  @apply z-1;
+  @apply rounded-l-none;
 }
 
 /* hide ie specific input controls */
--- a/assets/input/basic.css
+++ b/assets/input/basic.css
@@ -51,7 +51,6 @@ textarea:focus {
   @apply border-teal outline-none;
 }
 
-
 input[type="number"]:hover,
 input[type="text"]:hover,
 input[type="checkbox"]:hover:not(.checkbox-buttonStyle),
@@ -86,6 +85,25 @@ input[type="number"][data-valid="false"]:focus,
 input[type="text"][data-valid="false"]:focus,
 textarea[data-valid="false"]:focus {
   @apply text-grey-dark;
+}
+
+input.input_left {
+  @apply rounded-l-4 rounded-r-none;
+}
+
+input.input_right,
+div.input_right {
+  @apply rounded-r-4 rounded-l-none -ml-1;
+}
+
+.input_left:hover,
+.input_right:hover {
+  @apply z-1 transition duration-200;
+}
+
+.input_left:focus,
+.input_right:focus {
+  @apply z-1;
 }
 
 /* hide ie specific input controls */
--- a/src/__tests__/__snapshots__/node.test.ts.snap
+++ b/src/__tests__/__snapshots__/node.test.ts.snap
@@ -661,6 +661,8 @@ Object {
     ],
     "zIndex": Array [
       "responsive",
+      "hover",
+      "focus",
     ],
   },
 }
--- a/src/__tests__/__snapshots__/node.test.ts.snap
+++ b/src/__tests__/__snapshots__/node.test.ts.snap
@@ -252,6 +252,7 @@ Object {
       },
     },
     "margin": Object {
+      "-1": "-1px",
       "-headerMenu": "-14px",
       "-headerMenuMore": "-28px",
       "-selectMenu": "-21px",
@@ -400,6 +401,7 @@ Object {
       "scrollbar": "14px",
     },
     "zIndex": Object {
+      "1": 1,
       "auto": "auto",
       "card": 10,
       "clearButton": 10,
--- a/src/components/dropdown/withAutosuggest.tsx
+++ b/src/components/dropdown/withAutosuggest.tsx
@@ -269,19 +269,26 @@ function DropdownWithAutosuggest<T>({
       toggle={renderToggle}
       menu={(downshift, filteredOptions) => {
         return (
-          <Menu<T>
-            {...{
-              ...downshift,
-              innerRef: menuRef,
-              options: filteredOptions,
-              //className: "shadow-soft rounded-4",
-              equal: equalWrapper,
-              noResults,
-              hasHint,
-              inputValue,
-              isFetching,
-            }}
-          />
+          <div
+            className={classNames("", {
+              "-mt-selectWithHintMenu": hasHint,
+              "-mt-selectMenu": !hasHint,
+            })}
+          >
+            <Menu<T>
+              {...{
+                ...downshift,
+                innerRef: menuRef,
+                options: filteredOptions,
+                //className: "shadow-soft rounded-4",
+                equal: equalWrapper,
+                noResults,
+                hasHint,
+                inputValue,
+                isFetching,
+              }}
+            />
+          </div>
         )
       }}
     />
--- a/src/components/dropdown/withAutosuggest.tsx
+++ b/src/components/dropdown/withAutosuggest.tsx
@@ -63,8 +63,8 @@ interface Props<T> {
   fetchSuggestions?: (
     value: string
   ) => Promise<Array<{ value: T; name: string }>>
-  menuClassName?: string
   noResults?: string
+  hasHint?: boolean
 }
 
 const filterOptions = (noResults) => (allOptions, text) => {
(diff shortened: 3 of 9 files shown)


talamcol:
@masone I played around a lot and tried different things based on our discussion today. I was not able to find a nice solution.

However, I think the consistency improvement with the shadow and removing className on input and select give some value. Is it fine if I proceed and ignore the menuClassName?

assets/input/basic.css
```diff
@@ -88,6 +87,25 @@ textarea[data-valid="false"]:focus {
...
+
+.input_left:hover,
+.input_right:hover {
+  @apply z-1 transition duration-200;
+}
+
+.input_left:focus,
+.input_right:focus {
+  @apply z-1;
+}
```
masone, line 107: all of these look like they could be done with pure tailwind classes. can you try that please? as css modules are loaded with the js chunk, we try to use it only when necesssary. most of the values here look that they're likely to be reused somewhere else, so it should have little impact on the global css size

masone: as always when changing z-indexes, please give everything that might be overlapping these elements an explicit test run. for example, open a modal on top of it, see if tooltips are never overlapping, check the dropdown menu in the header, etc. (maybe instruct QA to test this as well)

src/__tests__/__snapshots__/node.test.ts.snap
```diff
@@ -400,6 +401,7 @@ Object {
       "scrollbar": "14px",
     },
     "zIndex": Object {
+      "1": 1,
```
masone, line 404: in the margin, we use `px` as the key for `1px`

src/__tests__/__snapshots__/node.test.ts.snap
```diff
@@ -252,6 +252,7 @@ Object {
       },
     },
     "margin": Object {
+      "-1": "-1px",
```
masone, line 255: in the paddings, we use `negative` for the key for `-1px`. I'm open to any variation, but would welcome consistency

src/__tests__/__snapshots__/node.test.ts.snap
masone: can you please check if this value is in the shared config already?

src/__tests__/__snapshots__/node.test.ts.snap
masone: can you please check if this value is in the shared config already?

src/components/input/index.tsx
```diff
@@ -64,7 +66,7 @@ const Input = forwardRef<HTMLInputElement, Props>(
       step,
       min,
       max,
-      className,
+      position,
```
masone, line 69: much better! 🙇

masone (changes requested):
Thanks for the good description, taking a structured approach and for weighting tradeoffs. This is a good improvement. Even when there are still classnames left, I would consider closing the ticket after this. We can always tackle it, should it ever become an issue

I have some small comments before merging, but am overall very happy with the approach you took! 🙏

assets/input/basic.css
talamcol: will try to use tailwind classes. But just as a sidenote regarding testing. The CSS here is a 1:1 copy from dealerhub and listings. It was duplicated there.
But I'll test again after I applied the tailwind classes

talamcol: The problem here is that rounder-r-none/l-none is overwritten by the global styling of our input component. It did not work to apply the classes in the component itself

Regarding testing, the CSS is a copy from dealerhub and listings (it was duplicated there). But I'll test it again 👍

src/__tests__/__snapshots__/node.test.ts.snap
talamcol: > can you please check if this value is in the shared config already?

in what package is the shared tailwind config included? I checked https://github.com/carforyou/carforyou-configuration-pkg but there is no tailwind configuration.

talamcol: > in the margin, we use `px` as the key for `1px`

As far as I know, z-index has no `px` value?

src/__tests__/__snapshots__/node.test.ts.snap
talamcol: > in the paddings, we use `negative` for the key for `-1px`. I'm open to any variation, but would welcome consistency

I think we are using `negative` only for the z-index key and not the padding. I checked how it is done in dealerhub and listings and there it is `"-1": "-1px"`
Should I change them to use negative as the key?

assets/input/basic.css
masone: I see. on class-level, the specificity is defined by the order of occurrence in the resulting stylesheet. These selectors might work around exactly that. We should still try reducing the amount of custom CSS as much as possible. I'm open to a compromise

Please make sure after your change, there is no duplicated CSS anymore :)

src/__tests__/__snapshots__/node.test.ts.snap
masone: 1. Yeah, your z-index naming makes sense. I didn't think hard enough

2. Sorry for causing confusion. I meant to reference to the [shared tailwind config in components-pkg](https://github.com/carforyou/carforyou-components-pkg/blob/master/src/tailwind/defaultConfig.ts). It is merged with project-/package-specific values here https://github.com/carforyou/carforyou-header-footer-pkg/blob/master/tailwind.config.js
Currently in this package, it seems there are no custom values.

src/__tests__/__snapshots__/node.test.ts.snap
masone: generally, I would push for as much consistency as possible. I'm wondering whether we need to define specific `-`/negative keys at all. I would assume tailwind generates them from the regular ones, for example `-mt-5` and `mt-5` when there's a `"5": "5px"`

masone: that said - it depends on your capacity. if you'd like to clean up, please do. if you feel it's a rabbit hole that's not worth it, try something sensible in the context you're in. it's nice to continually clean up on the side, but it should not prevent us from shipping stuff

tl;dr: as long as we don't make it harder to mantain, I'm fine with finding that there's something we can optimize

assets/input/basic.css
talamcol: done - removed everything except `rounded-l/r-none` from the global css

src/__tests__/__snapshots__/node.test.ts.snap
talamcol: I tested it, `-` is required on custom values. please see here for reference: https://tailwindcss.com/docs/margin#negative-values

But in general, I think the custom `-1` is not needed along with other custom values we defined in the config because tailwind already defines those. And I think as soon as we overwrite them, the functionality that you can use +/- is gone and you have to define both.

I can try to remove them but I'd prefer to create another ticket so that it gets tested well. Is this ok?

talamcol:
FYI: I'll mark that commit as BREAKING CHANGE when I merge it and will integrate the new version in the two related PR's from listing and dealerhub

src/__tests__/__snapshots__/node.test.ts.snap
masone: thanks for confirming. I was off on this :)

can absolutely be a separate thing. just this one would be a tiny improvement only. would only tackle it if we accidentally duplicated some values with different keys. after all, this will generate global css, which turns out to be very hard to reduce

masone (approved):
looking slick! congrats on the achievement!

consider giving QA detailed instructions on how to test this, ie. some components where this is used. because of z-index - keep an eye out for overlapping elements like dropdowns, tooltips and modals

src/components/dropdown/withAutosuggest.tsx
```diff
@@ -272,7 +272,7 @@ function DropdownWithAutosuggest<T>({
               ...downshift,
               innerRef: menuRef,
               options: filteredOptions,
-              className: classNames("shadow-soft rounded-4", menuClassName),
+              className: menuClassName,
```
GutierrezGo, line 275: It looks much better. No more overwriting styles. Thanks for checking it with a designer.

GutierrezGo (approved):
🍻

src/__tests__/__snapshots__/node.test.ts.snap
talamcol: Will check all of them as my next [task](https://autoricardo.atlassian.net/browse/CFYEO-40)

PR #298, opened 2021-01-11
fix: extract header footer tailwind config

ufkaya:
BREAKING CHANGE

the new configs are in this PR https://github.com/carforyou/carforyou-header-footer-pkg/pull/81


src/__tests__/__snapshots__/node.test.ts.snap
```diff
@@ -389,12 +383,9 @@ Object {
       "auto": "auto",
       "checkbox": "24px",
       "clearButton": "32px",
-      "logoDefault": "170px",
```
Averethel, line 392: I'm not 100% sure logo styles should be extracted. This way if you want to use the logo at the right size you need to include the header package which is not the most intuitive.
Is there a chance we can provide logo component here?

ufkaya: in case we need to use the logo somewhere else, we can extract them or use other values, these values are for the header

lkappeler (approved):
LGTM

masone:
ping! this seems valuable and it has been approved some while ago. let us know if you'd like support for merging it :)

ufkaya:
> ping! this seems valuable and it has been approved some while ago. let us know if you'd like support for merging it :)

I need to release it together with the new header/footer.
I think we didnt extract the styles because we cant have many different tailwind configs, the ones that I added in the header/footer pkg, are not visible on CFY and DH, and this is because I added them in the file `tailwind.config.js` which is valid only for the current pkg, right?
I will move the styles here again 😄 wdyt?

masone:
Damn. That indeed was the shortcut we've made initially and that's the reason these values were in components instead of header-footer. Same as you, I find that utterly confusing. Here is what the right thing to do would be:

1. header-footer-pkg would need to build and expose it's own tailwind config. probably only the custom one, not the merged one
2. the project would then need to merge the configs from components plus header-footer with it's own. preferrably, the `withDefaultConfig` in components-pkg would take multiple configs at once

it's not a crazy effort and it would certainly be the right thing to do, but given the work we have ahead, I'll leave it up to you to call the shots on how you want to go about it. to speed up, we could also pair on it

masone:
@ufkaya what did you end up doing regarding the issues you mentioned above? it sounded like this approach would not work, but now we merged it anyway

ufkaya:
thanks for reminding me, forgot to write the update.
We decided to extract them and use them as custom CSS on the header-footer pkg

masone:
Ah, I remember, thanks! always good to have it written down :)

PR #300, opened 2021-01-14
CFYEO-28 : extract Garage Promotion Card Component

Marine-Berthier:
Reference [CFYEO-28](https://autoricardo.atlassian.net/browse/CFYEO-28?atlOrigin=eyJpIjoiZjE1ODdlNDllMjkwNGNjZTg5YmNjZmYyNzIxZGI4YzQiLCJwIjoiaiJ9)

FixedDealerPromotion in listings project and PreviewCard in DH were similar components so It has been decided to extract this component to the component-pkg. 

In this PR you will find the GaragePromotionCard component which aims at replacing FixedDealerPromotion and PreviewCard. There is as well the related StoryBook page.


src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+  linkToDealerPage,
+  handleOnClick,
+  handleOnError,
+  handleOnClickText,
+}) => {
+  return (
+    <FixedCard
+      renderImage={() => (
+        <div>
+          {!mainImage.includes("null") ? (
```
ufkaya, line 40: Are you sure that the value is a string? Otherwise `!mainImage` is enough

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+            <img
+              src={mainImage}
+              className="promotionImage w-12/12 object-cover"
+              alt={title}
+              onClick={() => {
+                handleOnClick
```
ufkaya, line 46: This is an optional prop, you need to check if is available before you call it, and btw, you forgot to add `()` -> `handleOnClick()`

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+              }}
+              onError={() => {
+                handleOnError
```
ufkaya, line 49: same here

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+          ) : (
+            <UploadImagesIcon width="72px" height="72px" />
+          )}
+        </div>
+      )}
+    >
+      <div
+        className="p-20 w-12/12 md:w-card max-h-dropdown"
+        onClick={() => {
+          handleOnClickText
```
ufkaya, line 61: same here

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+          target="_blank"
+          rel="noreferrer"
+          className="text-teal flex items-center font-bold mt-16"
+        >
+          {previewLabel}
+          <ArrowRightTeal width="32px" height="32px" />
+        </a>
+        <hr className="text-grey-1 my-20" />
+        <div className="w-12/12 flex items-center">
+          {!logo.includes("null") ? (
```
ufkaya, line 85: we never use `null` as a string, can you check?

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+            <div className="w-2/12 mr-10">
+              <img src={logo} className="flex-shrink-0 mr-10" alt="logo" />
+            </div>
+          ) : null}
+          <div className="w-10/12">
+            <div className="font-bold">{dealerName}</div>
+            <div className="text-sm mt-5">{`${dealerLocation.address}, ${dealerLocation.zipCode} ${dealerLocation.city}`}</div>
```
ufkaya, line 92: `address` and `city` are optional, and if we dont pass them then the result is something like this `, 8001 `

src/components/garagePromotionCard.tsx
Marine-Berthier: On listings it looks like this : ``logo={`${cdnUrl}/${dealerPromotion.promotion.logo}`}``. Since the full path is passed at the component level I am not sure how I can check only the s3key within the component. There might be a better way to do it though.

ufkaya: I see, maybe you can check on the listings if the logo is available: `logo={dealerPromotion.promotion.logo ? `${cdnUrl}/${dealerPromotion.promotion.logo}` : null}`,

assets/card.css
```diff
@@ -7,3 +7,7 @@
 .fixedCard:hover {
   @apply shadow-soft transition duration-200;
 }
+
+.promotionImage {
+  height: 209px;
+}
```
masone, line 13: this value seems to be only used for that one component. in that case, we'd prefer to use a css module. if at all possible, we try not to rely on global styles, see https://github.com/carforyou/carforyou-docs/blob/master/frontend/css-conventions.md#2-css-modules

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+              alt={title}
+              onClick={() => {
+                handleOnClick
+              }}
+              onError={() => {
+                handleOnError
+              }}
+            />
+          ) : (
+            <UploadImagesIcon width="72px" height="72px" />
```
masone, line 53: displaying an upload image is a DH-specific requirement. can this be just passed as the main image instead of adding logic?

I like the "principle of least surprise" and looking at the props, it would definitely surprise me that there was some implicit image being displayed when I pass none

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
+import React, { FC, MouseEvent } from "react"
```
masone, line 1: this comment is regarding the filename - do we imagine ever using this component for something else than for the garage promotion? then I'd rename it to something more generic already now

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+import ArrowRightTeal from "./icons/arrowRight"
+import { FixedCard } from "./card/fixed"
+
+interface Props {
+  dealerName: string
+  dealerLocation: {
+    address?: string
+    zipCode: string
+    city?: string
+  }
```
masone, line 13: if you go for a more abstract name, I'd also go for more abstract names here. don't focus on the `dealer` so much - this is a specific detail of how the component is currently used. imagine we want to display the same card, but for a blog teaser for example. then it would be more like...

image, title, link, href, footerTitle, footerText (or similar)

src/components/garagePromotionCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+  title?: string
+  previewLabel: string
+  mainImage?: string
```
masone, line 16: are there any secondary images? or can this just be called `image`?
I would consider making this mandatory too. it will not look any good without an image I'd say...

masone (commented):
The component is currently very specific to the garage promotion - which is fair because we use exactly this in DH and listings, so the exact extraction is of value. I would support you to generalize even more, but I'm open to continuing as-is

src/components/garagePromotionCard.tsx
Marine-Berthier: Yes you're right. I will find more generic names for both  file and props.

src/components/garagePromotionCard.tsx
Marine-Berthier: No secondary image. I will make it mandatory 👍

src/components/icons/arrowRight.tsx
```diff
@@ -0,0 +1,15 @@
...
+      stroke="#3696B9"
+      strokeWidth="1.8"
+      d="M13 10l6 6-6 6"
+      fill="none"
+      strokeLinecap="round"
+    />
+  </svg>
+)
+
+export default ArrowRightTeal
```
masone, line 15: we try to have as little svgs as possible. if possible, can you use the existing arrow and use css to style the text and rotation? something like...
```
<ArrowDownM className="text-teal rotate-180 transform" />
```

src/components/previewCard/index.module.css
```diff
@@ -0,0 +1,3 @@
+.promotionImage {
+    height: 209px;
+  }
```
masone, line 3: some inconsistent formatting here. indent with 2 spaces and don't indent the closing bracket (not sure why the style linter does not help you with that)

src/components/previewCard/index.tsx
```diff
@@ -0,0 +1,107 @@
+import React, { FC } from "react"
+
+import styles from "./index.module.css"
+import UploadImagesIcon from "../icons/uploadImage"
+import ArrowRightTeal from "../icons/arrowRight"
+import { FixedCard } from "../card/fixed"
```
masone, line 6: can you expose the FixedCard as a default export instead of a named one please?

src/components/previewCard/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  handleOnClick,
+  handleOnError,
+  handleOnClickText,
+}) => {
+  return (
+    <FixedCard
+      renderImage={() => (
+        <div className="bg-grey-1 flex justify-center items-center text-grey-2 md:h-cardImage w-12/12 md:w-card">
+          {!image && editMode ? (
+            <UploadImagesIcon width="72px" height="72px" />
```
masone, line 44: Instead of relying on two boolean props, I would prefer to pass the image with `image` explicitly

src/components/previewCard/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+          handleOnClickText && handleOnClickText()
+        }}
+      >
+        {title ? (
+          <p className="text-lg">{title}</p>
+        ) : (
+          <>
+            <div className="bg-grey-1 h-20 w-12/12" />
+            <div className="bg-grey-1 h-20 w-12/12 my-10" />
+            <div className="bg-grey-1 h-20 w-6/12" />
```
masone, line 73: what are these 3 empty divs doing?

src/components/previewCard/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <div className="w-2/12 mr-10">
+              <img src={logo} className="flex-shrink-0 mr-10" alt="logo" />
+            </div>
+          ) : null}
+          <div className="w-10/12">
+            <div className="font-bold">{name}</div>
+            {location.address ? (
+              <div className="text-sm mt-5">{`${location.address}, ${location.zipCode} ${location.city}`}</div>
+            ) : (
+              <div className="text-sm mt-5">{`${location.zipCode} ${location.city}`}</div>
```
masone, line 98: duplicated styles in both branches of the condition. could we just wrap the condition instead?

src/components/previewCard/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  }
+  title?: string
+  previewLabel: string
+  image?: string
+  logo?: string
+  link?: string
+  editMode?: boolean
+  handleOnClick?: () => void
+  handleOnError?: (e) => void
+  handleOnClickText?: () => void
```
masone, line 23: for my understanding - when do `handleOnClick` and `handleOnClickText` differ?

src/components/previewCard/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    city?: string
+  }
+  title?: string
+  previewLabel: string
+  image?: string
+  logo?: string
+  link?: string
+  editMode?: boolean
+  handleOnClick?: () => void
+  handleOnError?: (e) => void
```
masone, line 22: does `handleOnError` differ depending on where the component is used?

src/components/previewCard/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+import ArrowRightTeal from "../icons/arrowRight"
+import { FixedCard } from "../card/fixed"
+
+interface Props {
+  name: string
+  location: {
+    address?: string
+    zipCode: string
+    city?: string
+  }
```
masone, line 14: instead of locking this component in to accepting a location, we could also have it accept a generic `footerText` or similar

masone (commented):
Nice iteration!! In components-pkg, we're always looking to make things as reusable as possible. We also tend to be a bit stricter about getting the interface right because changing away from a shared interface is always harder (breaking changes require us to update multiple projects)

let me know if you don't have the energy to push further. if you disagree with some of the comments or want to clear something up, let's schedule a call

src/components/previewCard/index.tsx
Marine-Berthier: It was used in DH, when there's no title it displays grey lines. From what I understood it was like a placeholder. I can add an if statement to show it only in DH ?

src/components/previewCard/index.tsx
Marine-Berthier: These are used on Listings for gtmHelper : the label for the first one is "click image" and the label for the second one is "click text" 😓

src/components/previewCard/index.tsx
Marine-Berthier: Currently it is used only on listings (where the components is used just once)

src/components/previewCard/index.tsx
masone: I see. generally, the component should be agnostic to where it's used and it should not know any details of the project where it's placed. so, a conditional to check which project it is used in would be a breach of this principle

if the content in a shared component is supposed to be flexible, [render props](https://reactjs.org/docs/render-props.html) are often a good choice. the DH-specific markup would be passed in DH, whereas the listings-specific one would be passed in listings. this way, the component itself does not need to know about the details of what exactly the content looks like, but it reserves a specific place to include it

src/components/previewCard/index.tsx
masone: I would vote for simplifying this. pretty sure we don't care about it on such a level of detail 🤔

masone:
If you feel like pairing on this, give me a nudge. Sometimes it's way faster to look at the code in a synchronous call and call the shots together

src/components/previewCard/index.tsx
Marine-Berthier: I get it ! I will drop `editMode` and try with render props

src/components/previewCard/index.tsx
masone: `onImageFailedLoading`?

src/components/previewCard/index.tsx
masone: we discussed to drop the `editMode` prop and just rely on either passing an image or not. as listings will always have an image, it should never show this upload icon

we didn't really find a viable solution to take everything apart nicely

src/components/previewCard/index.tsx
masone: we discussed to keep it as is
we didn't really find a viable solution to take everything apart nicely

masone:
Marine and I talked for an hour and explored how to make this component nice and generic. Unfortunately we did not find a viable way. We decided to live with a somewhat not-so-generic component that contains more business logic as we usually like it to have (eg. the placeholders and the upload icon). it's just too hard to generalize it the way it behaves now. 

extracting this will still be a win for sharing it between the two projects

src/components/previewCard/imageHelper.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { FC } from "react"
+
+import styles from "./index.module.css"
+
+interface Props {
+  imageSrc?: string
+  pictureAlt?: string
```
ufkaya, line 7: We should render this component only if we have the imageSrc, so `imageSrc` and `pictureAlt` should be mandatory

src/components/previewCard/imageHelper.tsx
```diff
@@ -0,0 +1,27 @@
...
+  handleOnErrorImage?: (e) => void
+}
+
+const ImageHelper: FC<Props> = ({ imageSrc, pictureAlt, handleOnErrorImage }) =>
```
ufkaya, line 11: Can you rename it to just `Image`, usually we use `helper` for libraries

src/components/previewCard/imageHelper.tsx
```diff
@@ -0,0 +1,27 @@
...
+  imageSrc ? (
+    <img
+      src={imageSrc}
+      alt={pictureAlt}
+      className={
+        pictureAlt == "logo"
```
ufkaya, line 17: can you try to use a new prop instead of the `pictureAlt` value?

Marine-Berthier: @ufkaya something like `isLogo` ?

ufkaya: Yes

src/components/icons/uploadImage.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from "react"
```
masone, line 1:
```suggestion
import React from "react"
```

masone: I _think_ importing `*` takes more than necessary and might have an effect on the resulting bundle (not sure here). Just importing `React` only works fine

src/components/previewCard/imageHelper.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface Props {
+  imageSrc: string
+  pictureAlt: string
+  isLogo: boolean
+  handleOnErrorImage?: (e) => void
+}
+
+const Image: FC<Props> = ({
+  imageSrc,
+  pictureAlt,
```
masone, line 14:
```suggestion
  imageAlt,
```

otherwise I'm thinking that a `picture` is something different from an `image` :)

src/components/previewCard/imageHelper.tsx
```diff
@@ -0,0 +1,32 @@
...
+import styles from "./index.module.css"
+
+interface Props {
+  imageSrc: string
+  pictureAlt: string
+  isLogo: boolean
+  handleOnErrorImage?: (e) => void
+}
+
+const Image: FC<Props> = ({
```
masone, line 12: please rename the file to `image.tsx`, so it's in line with the naming of the component

src/components/previewCard/imageHelper.tsx
```diff
@@ -0,0 +1,32 @@
...
+    alt={pictureAlt}
+    className={
+      isLogo
+        ? "flex-shrink-0 mr-10"
+        : `${styles.promotionImage} w-12/12 object-cover`
+    }
+    onError={(e) => {
+      handleOnErrorImage && handleOnErrorImage(e)
+    }}
+  />
```
masone, line 29: I would totally duplicate that `img` tag and not extract it. now we just need a condition `isLogo`. if it was not extracted, the styles would just be inlined where they're used

Marine-Berthier: Yes I dropped the "imageHelper", it is more straightforward this way.

src/components/previewCard/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <div className="w-12/12 flex items-center">
+          {logo ? (
+            <div className="w-2/12 mr-10">
+              <img
+                src={logo}
+                alt={"logo"}
+                className={"flex-shrink-0 mr-10"}
+                onError={(e) => {
+                  onImageFailedLoading && onImageFailedLoading(e)
+                }}
```
masone, line 78: We will not be able to apply the same fallback behavior to the large and the small image  alike. I suggest to simply ignore a failing logo and only allow to create a fallback for the large image

```suggestion
```

masone (approved):
i'm really happy about the end result of this. it looked like we would not find some sensible solution, but you really managed to optimize to the max here. thanks for the patience! 

Please drop the `onError` from the logo and then you're good to merge from my side!
Remember to apply a semantic commit message when squash-merging

PR #301, opened 2021-01-19
feat: checkbox filter

Averethel:
Extracts a slightly simplified version of checkbox filter component from listing, that we could use for dealer-hub filters.

JIRA: https://autoricardo.atlassian.net/browse/CAR-6011

Screenshot 2021-01-19 at 15 30 00

Screenshot 2021-01-19 at 15 29 54

Screenshot 2021-01-19 at 15 29 47

Screenshot 2021-01-19 at 15 29 43

Screenshot 2021-01-19 at 15 29 38


ralucacril (approved):

src/components/filters/__tests__/checkbox.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+      const { container } = renderWrapper({
+        facet: { "1": 30, "2": 1000, "3": 0 },
+      })
+
+      expect(container).toMatchSnapshot()
+    })
+
+    it("with title", () => {
+      const { container } = renderWrapper({
+        children: <div>Title</div>,
```
lkappeler, line 98: Why should the title be a child of the component, couldn't we just render it on top?

src/components/filters/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const selectionCopy = [...selected]
+
+    if (selectionCopy.includes(value)) {
+      const i = selectionCopy.indexOf(value)
+      selectionCopy.splice(i, 1)
+    } else {
+      selectionCopy.push(value)
+    }
+
+    selectionCopy.sort((a, b) =>
```
lkappeler, line 41: Why do they need to be sorted?

src/components/filters/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+      <div className="w-12/12 -mb-5" key={`filter_${name}_${value}`}>
+        <Checkbox
+          name={`filter_${name}_${value}`}
+          renderLabel={() => (
+            <span className="text-base leading-xs flex items-center">
+              {renderIcon ? renderIcon() : null}
+              {label}
+              {typeof resultsCount === "number" && (
+                <span className="inline-block ml-5 text-grey-3 text-sm">
+                  ({formatNumber(resultsCount)})
```
lkappeler, line 63: I would not expect the component to format the number for me, rather that renders what I pass...

src/components/filters/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+                </span>
+              )}
+            </span>
+          )}
+          labelPosition="left"
```
lkappeler, line 68: I would probably make that an option so the component is more flexible/generric

src/components/filters/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+  options: Item[]
+  applyFilters: (filtersToApply) => void
+  selected?: (string | number)[]
+  facet?: { [key: string]: number }
+}
+
+const CheckboxFilter: FC<Props> = ({
+  name,
+  options,
+  applyFilters,
```
lkappeler, line 24: maybe call it `onChange`for consistency?

ufkaya (approved):

src/components/filters/__tests__/checkbox.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+      })
+
+      expect(container).toMatchSnapshot()
```
masone, line 101: I would prefer to assert on the title being present over using snapshots allover the place :)

src/components/filters/__tests__/checkbox.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+            label: "Three",
+            value: 3,
+            renderIcon: () => (
+              <ClockOutlinedTealIcon width="20" height="20" className="mr-10" />
+            ),
+          },
+        ],
+      })
+
+      expect(container).toMatchSnapshot()
```
masone, line 79: I would prefer to assert on the icon being present over using snapshots allover the place. maybe add a test-id to the icon you're passing and get that with react-testing-lib? :)

src/components/filters/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+    applyFilters({ [name]: selectionCopy })
+  }
+
+  const renderOption = ({ value, label, renderIcon }: Item) => {
+    const isSelected = selected.includes(value)
+    const resultsCount = facet[value.toString()]
+    const isDisabled = resultsCount === 0 && !isSelected
+
+    return (
+      <div className="w-12/12 -mb-5" key={`filter_${name}_${value}`}>
```
lkappeler, line 54: What is the negative margin used for?

src/components/filters/checkbox.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+interface Props {
+  name: string
+  options: Item[]
+  applyFilters: (filtersToApply) => void
+  selected?: (string | number)[]
+  facet?: { [key: string]: number }
+}
+
+const CheckboxFilter: FC<Props> = ({
```
lkappeler, line 21: I would update the interface to be a bit more generic and not having it fit 1:1 the current implementation.
Something like the following so the option would hold the information and we could ditch the update logic and just update the option that is passed and return it to the callback.

```
interface Item {
  value: string | number
  label: string | () => ReactNode // So it could be flexible to display the icon at the start or end, to have multiple icons etc.
  selected: boolean
  count: number
  labelPosition?: LabelPosition
}

interface Props {
  name: string
  options: Item[]
  onChange: (options) => void
}

```

masone (approved):
LGTM! India has a PR open for replacing the color filter and they're struggling getting the change out as-is. Please be considerate with the timing of introducing this in listings to not impose unnecessary merge conflicts on their pending work

src/components/filters/__tests__/checkbox.test.tsx
Averethel: Makes sense, I was lazy and copied existing tests

src/components/filters/__tests__/checkbox.test.tsx
Averethel: Makes sense, I was lazy and copied existing tests

lkappeler (commented):
A couple of thought on the interface we might benefit form in the future

Averethel:
@masone I'm not 100% sure if we would introduce it in listings now. My take on it is that we only do it if we have the capacity and more towards the endo of the quarter, but let's see how we are progressing.

src/components/filters/__tests__/checkbox.test.tsx
Averethel: Fair, I suppose we can just drop it.

src/components/filters/checkbox.tsx
Averethel: Consider the following:

1. 
  - [ ] A
  - [x] B


2. 
  - [x] A
  - [x] B


3. 
  - [x] A 
  - [ ] B

Whan unsorted  going 1-2-3 would result in state changes `[]` -> `[B]` -> `[B, A]` -> `[A]`, while 3-2-1 would result in `[]` -> `[A]` -> `[A, B]` -> [B]` so the both filters applied would result in a different state. Given that this state can be used in various ways (e.g. to generate URL) I want to not care about the order the filter s are applied.

src/components/filters/checkbox.tsx
Averethel: Let's cross that bridge when we get there. We haven't had the need for that so far

src/components/filters/checkbox.tsx
Averethel: To reduce the space that comes from the checkbox.

lkappeler: Isn't that breaking the space needed to display a potential validation error then?

Averethel: It could if single checkbox could ever show the validation error. That is not the case though.

src/components/filters/checkbox.tsx
lkappeler: I do get that the order of the selected array would vary based on the way the way the filters are applied I just wondered why it would matter for the application if we have something like `?filter=A,B` as a query wouldn't we get the same result if the query is `?filter=B,A`

Marine-Berthier (approved):

src/components/filters/checkbox.tsx
Averethel: In terms of content - yes, in terms of things like browser caching - no since the URL would be different.

src/components/filters/checkbox.tsx
Averethel: Fair, this allows us to not duplicate the formatting helper

src/components/filters/checkbox.tsx
Averethel: I'll leave it open for now and see how it feels when using it. This looks like an excellent interface for a render prop (to render a single option), but since filters are stateless my concern is that you end up deconstructing and reconstructing the selection all over the place. Potentially this could be refactored later on.

lkappeler (approved):
@Averethel Feel free to merge back when you feel comfortable, my comments are not blocking just some thought I had while reviewing :)

Averethel:
@lkappeler Thanks, I figured since you did not request changes. I want to use the component in the context of dealer-hub filters to see if there weren't anything that I missed or whether I want to fiddle with the interface.

masone:
ping :) can I help to move this forward? looks like it's ready to merge!

Averethel:
I want to merge it after implementing first filters in dealer hub, this way there is a bit more wiggle room to adjust the interface.

PR #303, opened 2021-01-27
CAR-5804: write stories in tsx

mreljin:
- Refactored the stories from .js to .tsx.


lkappeler:
@mreljin Hey, I think we run into some issues with the pre releases for this repo due to the branch name.

Can you rename the branch so it does not contain chars not supported by semver, [0-9A-Za-z-] (like replace the underlines with dashes) as we build the name for the pre release tag based on the branch name, it will result in a invalid version if it contains chars other than the ones mentioned above..

See: https://semver.org/

PR #304, opened 2021-01-27
fix: make the onChange event compatible with default input on change event

lkappeler:
Makes the onChange event compatible with the default onChange event yielded by an input filed.

Use Case: We want to ad an input mask that formats the input for the user, the input mask library is submitting a `ChangeEvent<HTMLInputElement>` event, with the current typing the component can not be used as a wrapper ref: https://github.com/carforyou/carforyou-listings-web/pull/2861/files

### DIFF
--- a/src/components/input/index.tsx
+++ b/src/components/input/index.tsx
@@ -1,4 +1,9 @@
-import React, { FocusEvent, forwardRef, KeyboardEvent } from "react"
+import React, {
+  ChangeEvent,
+  FocusEvent,
+  forwardRef,
+  KeyboardEvent,
+} from "react"
 import classNames from "classnames"
 
 import InputField from "./inputField"
@@ -19,8 +24,16 @@ interface InputProps {
   hasClearButton?: boolean
   hasSearchIcon?: boolean
   mode: "text" | "numeric" | "decimal" | "tel" | "email"
-  onChange: <T extends { target: { name: string; value: string | number } }>(
-    e: T
+  onChange: (
+    e:
+      | ChangeEvent<HTMLInputElement>
+      | {
+          target: {
+            name: string
+            value: string | number
+            cleared: boolean
+          }
+        }
   ) => void
   onBlur?: (e: FocusEvent) => void
   onKeyDown?: (e: KeyboardEvent) => void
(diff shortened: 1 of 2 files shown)


src/components/input/index.tsx
```diff
@@ -19,8 +24,16 @@ interface InputProps {
...
-  onChange: <T extends { target: { name: string; value: string | number } }>(
-    e: T
+  onChange: (
+    e:
+      | ChangeEvent<HTMLInputElement>
+      | {
+          target: {
+            name: string
+            value: string | number
+            cleared: boolean
```
lkappeler, line 34: The reason we need the custom typing is that the component calls the onCange event with a cleared prop in the custom implementation to clear the form input. The `cleared` is non standard.

masone: can we extend the stock ChangeEvent with the additional prop instead of redefining it manually?

Averethel (commented):
Looks good. If we can achieve what Christian mentioned it would be great.

src/components/input/index.tsx
lkappeler: TLDR: Lets have a chat in it 🙂 

I tried this approach first the Issue I had is that the clear function is calling the on change event directly and not re-setting the value of the input directly so we miss some prototypes that are included when the event is fired by the browser. I checked the projects and we only use the cleared prop in two cases. But this would still require some refactoring. I tend towards dropping the cleared prop as the functionality can be repleced by checking if the new value is an empty string instead and we would have the standard event...

Averethel: Thanks for some explanation. I like the approach of ditching the custom thing, but I don't know if we have the capacity to do it now. Up to you to judge that.

Averethel (approved):

src/components/input/index.tsx
lkappeler: Yeah I already added it to the tech dept list, it's not a huge effort but I'd rather not release a breaking change and update the projects right now...

masone: instead of waiting for my answer, you called the shots plus documented the shortcoming. it's no biggie after all. thanks for handling it this way! that's how we're making most of the situation we're in right now 👍

PR #305, opened 2021-01-27
Car 5804: write stories in tsx

mreljin:
Refactored the stories from .js to .tsx.


mreljin:
@lkappeler, hey, thanks for the help! :) 
I saw the comment about the branch name only allowing [A-Za-z0-9-]

src/stories/dropdown.stories.tsx
```diff
@@ -45,7 +42,30 @@ export default {
...
+  onSelect?: (selection: T) => void
+  placeholder?: string
+  renderOption?: (option: {
+    value: T
+    name: string | ReactNode
+    isSelected: boolean
+  }) => ReactNode
+}
+
+const Template = <T extends Record<string, unknown>>(args: Props<T>) => {
```
masone, line 68: I have never used `Record`. can you outline what it generally does and why it's required here?

src/stories/dropdown.stories.tsx
```diff
@@ -45,7 +42,30 @@ export default {
...
+  toggle: (
+    selected: {
+      name: string
+      value: T
+      placeholder?: boolean
+    },
+    isOpen: boolean
+  ) => ReactChild
+  options: Array<{
+    value: T
```
masone, line 56: what are the possible types we might encounter for this generic type `T`?

src/stories/input/appearence.stories.tsx
```diff
@@ -51,7 +67,7 @@ WithClearButton.args = {
   value: "text",
   hasClearButton: true,
   placeholder: "Type something",
-  onChange: (e) => {
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => {
```
masone, line 70: would it make sense to move this type to the component and rely on type inference here instead?

src/stories/alertMessage.stories.tsx
```diff
@@ -24,7 +24,17 @@ export default {
...
-const Template = (args) => {
+interface Props {
+  storyName: string
+  text: string
+  children: ReactNode
+  icon: () => ReactNode
+  fullWidth?: boolean
+  type: "error" | "warning" | "information" | "success"
+  alignCenter?: boolean
+}
```
masone, line 35: I'm guessing without fact checking here - most of these are probably defined in the component itself. can we rely on extending the existing type of the component to prevent duplication?

for example, if we'd introduce a new type of alert message, we'd currently need to add it in two places

masone: (this applies to many stories below)

src/stories/dropdownWithAutosuggest.stories.tsx
```diff
@@ -44,7 +43,19 @@ export default {
...
 
-const Template = (args) => {
+interface Props<T> {
+  storyName: string
+  options: Array<{
+    value: T
+    name: string
+  }>
+  input: (propGetter: InputProps) => ReactChild
+  onSelect: (selection: T | { customValue: string }) => void
```
masone, line 53: can `customValue` be of type `T` or is it set in stone that it's a `string`?

src/stories/select/simple/label.stories.tsx
```diff
@@ -44,7 +57,7 @@ const Template = (args) => {
               <Select
                 {...args}
                 selected={value}
-                handleChange={(v) => {
+                handleChange={(v: React.ChangeEvent<HTMLSelectElement>) => {
```
masone, line 60: can this type be defined in the `handleChange` of the component instead?

src/stories/storyContainer.tsx
```diff
@@ -1,10 +1,16 @@
 import React from "react"
 
+interface StoryContainerArgs {
```
masone, line 3: for the sake of consistency - we usually name the interface of a component `Props` by default
```suggestion
interface Props {
```

masone (commented):
Nice, thanks! I'm pretty sure this will unveil some cases where we were violating the components interfaces. I have some questions on the generic `T` type for the values and am wondering whether we could prevent repeating the types of the components

src/stories/alertMessage.stories.tsx
mreljin: Yeah, that was my original idea, to extend the interface, but then I decided to just use the props that were actually used, not sure why. Thanks for the correction

src/stories/dropdown.stories.tsx
mreljin: I originally wanted T to extend an empty object:
const Template = <T extends {}>(args: Props<T>) => { <------ like this 

but npm run lint threw an error there:
Don't use `{}` as a type. `{}` actually means "any non-nullish value"

I'll cite stackoverflow, and link to some documentation since they are a bit better at explaining than me! :) 

> A Record<K, T> is an object type whose property keys are K and whose property values are T. That is, keyof Record<K, T> is equivalent to K, and Record<K, T>[K] is (basically) equivalent to T.

docs link: https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype

src/stories/alertMessage.stories.tsx
masone: thanks, this looks much cleaner and easier to maintain now

src/stories/modal.stories.tsx
```diff
@@ -55,7 +55,15 @@ function ModalDemo(args) {
   )
 }
 
-const Template = (args) => {
+interface Props {
+  storyName: string
+  label: string
+  size?: "small" | "medium" | "large" | "fullscreen"
+  style?: "white" | "dark"
+  onClose: () => void
```
masone, line 63: is it possible to extend the components props here as well or are the Props defined explicitly here on purpose?

src/stories/radioButton.stories.tsx
```diff
@@ -28,7 +28,12 @@ export default {
   },
 }
 
-const Template = (args) => {
+interface Props extends RadioButtonProps {
+  storyName: string
+  label: JSX.Element
```
masone, line 33: we keep repeating the definition for `storyName` and an optional `label` in every story we write. if we defined a type for it, we could enforce some common structure to our stories. wdyt? worth it?

masone: I'll let it up to you to decide. The conversion to tsx is worth so much on its own. I could totally live with some duplication here

masone (approved):
crazy cool! thanks for the extra run. I'll leave it up to you to decide how to proceed with the remaining comment. they shouldn't be blockers imo

src/stories/modal.stories.tsx
mreljin: The modal is not its own separate component for some reason, and it's defined within the `modal.stories.tsx` file instead of its own `modal.tsx` component file. So I'm the first to define Props for that.

image

I didn't want to touch the components as I'm still getting into the whole project and that just creates a risk of breaking something unintentionally.

src/stories/radioButton.stories.tsx
mreljin: Went the extra mile and did that. Check if it's okay

PR #306, opened 2021-02-01
feat: add datepicker

ufkaya:
- [CAR-6045](https://autoricardo.atlassian.net/browse/CAR-6045)

DatePicker is from the listings project


assets/datePicker.css
```diff
@@ -0,0 +1,117 @@
+/* purgecss start ignore */
```
ufkaya, line 1: I added a task to convert this css file to css module

lkappeler: @ufkaya What's the reason you can not use it the same way as in the listings project? https://github.com/carforyou/carforyou-listings-web/blob/master/components/forms/datePicker.module.css

lkappeler (commented):
@ufkaya I might have missed some discussions but whats the reason to duplicate it instead of moving it to the components packge?

assets/datePicker.css
ufkaya: tailwind style was not processed, and I didnt want to spend too much time now to fix it, but if you know how to solve the issue then I can change it again otherwise we can do it later, I added a task as tech debt

ufkaya:
> @ufkaya I might have missed some discussions but whats the reason to duplicate it instead of moving it to the components packge?

This is the component pkg, I will use it in the DH

lkappeler:
> This is the component pkg, I will use it in the DH

Oh lol nvm then 😂

assets/datePicker.css
```diff
@@ -0,0 +1,117 @@
...
+
+:root {
+  --tileSize: 47px;
```
lkappeler, line 4: Why did u add `--` to the variables?

lkappeler (approved):
Looks good to me

assets/datePicker.css
ufkaya: This is the syntax to  declare a CSS variable

masone: CSS variables are not supported by all browsers, while the previous approach was a css-modules feature (which afaik got transpiled)

masone (approved):
the additional global css is a fine tradeoff to take in this situation. thanks for handling it so transparently!

assets/datePicker.css
ufkaya: Do we still need to support IE?

ralucacril: nope, we do not support IE

ufkaya: thanks, then we are good https://caniuse.com/css-variables

lkappeler: Oh wasn't aware that a native support for this has been added...

PR #308, opened 2021-02-05
refactor: change global css to css modules

talamcol:
the date picker is currently using global css because tailwind values were not applied to the markup. 

- changed global css to css modules
- added a postcss configuration so that the tailwind values are correctly applied
- shared the config with the storybook


package.json
```diff
@@ -60,6 +60,7 @@
     "@types/node": "^12.12.24",
     "@types/react": "^16.9.17",
     "@types/react-paginate": "^6.2.1",
+    "autoprefixer": "^9.7.4",
```
ufkaya, line 63: Why do we need it? Isn't enough the one from tailwindcss?

ufkaya (approved):
LGTM, thanks for the fix!

package.json
talamcol: It is needed for the browser-specific prefixes such as `-moz`. We have the same in listings too. Maybe we can drop it in the future since we already dropped IE support

ufkaya: I know what it does, but is already required by tailwindCSS, therefore, it should be already installed, I think there is no need to install it separately, can you please check?

.storybook/postcss.config.js
```diff
@@ -1,9 +1,9 @@
...
-}
+const tailwindConfig = require("../.tailwind/defaultConfig").default
+
+module.exports = {
+  plugins: [
+    require("postcss-easy-import"),
+    require("tailwindcss")(tailwindConfig),
+    require("autoprefixer"),
+  ],
+}
```
masone, line 9: as this is exactly the same as the one in the root, you could try:

```
const config = require("../postcss.config")
module.exports = config
```

masone (approved):
looking good from my side! please verify it works in the project (maybe with a prerelease) and that storybook also still displays correctly :)

package.json
talamcol: I tested it, and the datepicker still looks fine. I'm not sure though if it affects anything else.
Based on this section, it only has the autoprefixer built-in if you use `postcss-preset-env`
https://tailwindcss.com/docs/using-with-preprocessors#future-css-features

I can remove it and see what happens?

masone: sure, go ahead. otherwise. you can also always give it to QA for cross-browser testing

ufkaya: Yes correct, we don't require postcss-preset-env but we only require and use the autoprefixer that's why it works and there is no need to install it separately.
We can also replace `require("autoprefixer") with require("postcss-preset-env")` so we will have a better support

talamcol: for some reason, I have to install it manually with this change. It removes the autoprefixer package from the node_modules otherwise (tested on my local and in the pipeline)

talamcol:
> looking good from my side! please verify it works in the project (maybe with a prerelease) and that storybook also still displays correctly :)

Tested storybook, listings, and dealerhub. All looked the same as before -> QA will test Ufuks PR, this change only gets merged into his branch.

PR #311, opened 2021-02-09
feat: Input filter

Averethel:
This PR extracts BaseInputFilter from listings so that we can reuse it in dealer-hub.
Entry for replacing in listings created in debt sheet


src/components/filters/__tests__/input.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+          name="test"
+          placeholder="test"
+          apply={jest.fn()}
+        />
+      )
+
+      expect(container).toMatchSnapshot()
+    })
+
+    it("with autofocus after focusing", () => {
```
masone, line 21: I don't even know what it means 🙈

src/components/filters/__tests__/input.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+          mode="numeric"
+          name="test"
+          placeholder="test"
+          apply={jest.fn()}
+        />
+      )
+
+      fireEvent.focus(getByPlaceholderText("test"))
+
+      expect(container).toMatchSnapshot()
```
masone, line 33: how is focussing changing the resulting snapshot? is there something specific we can expect instead of taking a snapshot?

src/components/filters/input.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  onFocus = (event) => {
+    event.persist()
+    this.setState({ refocus: true }, () => {
+      const closestSection = getClosestElement(
+        event.target,
+        "[data-closestpoint]"
+      )
+      scrollIntoViewIfMobile(closestSection)
```
masone, line 66: would it be possible to use a ref instead?

masone (approved):

src/components/filters/__tests__/input.test.tsx
Averethel: I believe it adds autofocus prop. Maybe @lkappeler can confirm that? I did copy over snapshots from listings...

src/components/filters/input.tsx
Averethel: Given this component is quite complex I would not refactor it too heavily in the scope of this PR. I'd be happy to tackle this separately.

src/components/filters/__tests__/input.test.tsx
Averethel: It adds `autofocus` prop to the input after the first focus event IIRC. 
The reason is that changing the URL results in the page (re)load and we don't want to lose focus when typing.

masone: I see. no need to change existing code when porting. I think it's a good example of how we can make our tests better in the future ;)

src/components/filters/input.tsx
masone: yeah, didn't realize you're just porting!

PR #313, opened 2021-02-16
BREAKING CHANGE: force collapse of a collapsible

Averethel:
We have a case where we need to be able to control the state of a collapsible from the outside.

An alternative to this approach would be to have a collapsible list component delivered by this package. It would then use a different item than the collapsible that is exported outside.

### DIFF
--- a/src/components/__tests__/collapsible.test.tsx
+++ b/src/components/__tests__/collapsible.test.tsx
@@ -7,12 +7,12 @@ const openedText = "Close me"
 const closedText = "Open me"
 const content = "I am content"
 
-const renderWrapper = ({ isInitiallyCollapsed = true, onChange = null } = {}) =>
+const renderWrapper = ({ forceCollapse = true, onChange = null } = {}) =>
   render(
     <Collapsible
       renderToggle={(isCollapsed) => (isCollapsed ? closedText : openedText)}
       onChange={onChange}
-      isInitiallyCollapsed={isInitiallyCollapsed}
+      forceCollapse={forceCollapse}
     >
       {() => content}
     </Collapsible>
@@ -42,7 +42,7 @@ describe("<Collapsible>", () => {
   describe("rendered open", () => {
     it("is open", () => {
       const { getByText, queryAllByText } = renderWrapper({
-        isInitiallyCollapsed: false,
+        forceCollapse: false,
       })
 
       expect(queryAllByText(closedText)).toEqual([])
@@ -52,7 +52,7 @@ describe("<Collapsible>", () => {
 
     it("can be closed", () => {
       const { getByText, queryAllByText } = renderWrapper({
-        isInitiallyCollapsed: false,
+        forceCollapse: false,
       })
 
       fireEvent.click(getByText(openedText))
--- a/src/components/collapsible.tsx
+++ b/src/components/collapsible.tsx
@@ -4,10 +4,6 @@ import classNames from "classnames"
 import ArrowDownM from "./icons/arrowDownM"
 interface Props {
   /**
-   * False if the element shall be expanded by default
-   */
-  isInitiallyCollapsed?: boolean
-  /**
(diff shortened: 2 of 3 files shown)


src/components/__tests__/collapsible.test.tsx
```diff
@@ -7,12 +7,12 @@ const openedText = "Close me"
 const closedText = "Open me"
 const content = "I am content"
 
-const renderWrapper = ({ isInitiallyCollapsed = true, onChange = null } = {}) =>
```
ralucacril, line 10: so from my understanding we will drop the isInitiallyCollapsed? aren't we using this in filters? 🤔  maybe you can explain a bit how forceCollapse is replacing the isInitiallyCollapsed cuz i am a bit confused

Averethel:
No, it will be replaced by `forceCollapse`. The difference is that when `forceCollapse` would change it would force the collapsible state to update. Whereas `isInitiallyCollapsed` was only responsible for setting the initial state.

I tired rolling with additional property and keeping it in sync with `isInitiallyCollapsed`, but it made the code insanely complex and hart to follow and allowed for some weird edge cases that would result in an impossible state (`forceCollapse: true` &`isInitiallyCollapsed: false` 🤯 )

src/components/__tests__/collapsible.test.tsx
Averethel: No, it will be replaced by `forceCollapse`. The difference is that when `forceCollapse` would change it would force the collapsible state to update. Whereas `isInitiallyCollapsed` was only responsible for setting the initial state.

I tried rolling with additional property and keeping it in sync with `isInitiallyCollapsed`, but it made the code insanely complex and hart to follow and allowed for some weird edge cases that would result in an impossible state (`forceCollapse: true` &`isInitiallyCollapsed: false` 🤯 )

ralucacril: aaa gotcha now. makes sense to roll just with forceCollapse. thanks for explaining it.

src/components/collapsible.tsx
```diff
@@ -25,16 +21,27 @@ interface Props {
...
   onChange,
   opacityOnHover = true,
+  forceCollapse = true,
 }) => {
-  const [isCollapsed, setIsCollapsed] = useState(isInitiallyCollapsed)
+  const [isCollapsed, setIsCollapsed] = useState(forceCollapse)
+
+  useEffect(() => {
+    setIsCollapsed(forceCollapse)
+  }, [forceCollapse])
```
ufkaya, line 44: I dont think that we need the useEffect here because we initiate the state with the prop forceCollapse and we never change it internally

src/components/collapsible.tsx
```diff
@@ -25,16 +21,27 @@ interface Props {
    * If the element should get transparent on hover
    */
   opacityOnHover?: boolean
+  /**
+   * Forces the collapsible to collapse
+   *   this can be used to allow control of the collapsed state outside of the component
+   *   an example use-case is controlling the focus of children collapsibles by the parent
+   *   component to ensure that only one is open at the time
+   */
+  forceCollapse?: boolean
```
lkappeler, line 30: Optional: maybe isCollapsed instead to keep it in in line with other state props

lkappeler (approved):

ralucacril (approved):
forgot to approve

src/components/collapsible.tsx
masone: I'm working on upgrading this in listings. not sure what the motivation was for introducing a breaking change for this... my two cents: the new name causes more confusion than the previous one

Averethel: @masone the motivation for the change was that we needed to be able to control multiple collapsibles from the parent component in the listing form in dealer-hub (so that we can close the ones that are not focused). Hence the change in behaviour and breaking change.

PR #316, opened 2021-02-22
feat: make segmented control grow

ralucacril:
https://autoricardo.atlassian.net/browse/CAR-6411

Currently all the segments are the same width, regardless of text content, which can lead to one segment being multiple lines. 
There are cases when we want to have dynamic width.

Before:
Screenshot 2021-02-22 at 12 18 10
Screenshot 2021-02-22 at 12 17 22

After:
Screenshot 2021-02-22 at 12 16 52


lkappeler (approved):
LGTM

PR #317, opened 2021-03-10
feat: add collapsible teal color chevron and mobile position chevron

acapop77:
feat: add collapsible teal color chevron and mobile position chevron

### DIFF
--- a/src/components/collapsible.tsx
+++ b/src/components/collapsible.tsx
@@ -30,6 +30,10 @@ interface Props {
   forceCollapse?: boolean
   /** Makes sure that the content is always in the DOM */
   alwaysRender?: boolean
+  /** Teal color chevron */
+  tealChevron?: boolean
+  /** Chevron position on the top for mobile and center for desktop */
+  mobileChevron?: boolean
 }
 
 const Collapsible: FC<Props> = ({
@@ -39,12 +43,19 @@ const Collapsible: FC<Props> = ({
   opacityOnHover = true,
   forceCollapse = true,
   alwaysRender = false,
+  tealChevron = false,
+  mobileChevron = false,
 }) => {
   const [isCollapsed, setIsCollapsed] = useState(forceCollapse)
 
   useEffect(() => {
     setIsCollapsed(forceCollapse)
   }, [forceCollapse])
+
+  const arrowColor = tealChevron ? "text-teal" : ""
+  const chevronPosition = mobileChevron
+    ? "items-start lg:items-center"
+    : "items-center"
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


masone:
can you link the specs / requirements please?

src/components/collapsible.tsx
```diff
@@ -39,13 +43,20 @@ const Collapsible: FC<Props> = ({
...
+  tealChevron = false,
+  mobileChevron = false,
 }) => {
   const [isCollapsed, setIsCollapsed] = useState(forceCollapse)
 
   useEffect(() => {
     setIsCollapsed(forceCollapse)
   }, [forceCollapse])
 
+  const arrowColor = tealChevron ? "text-teal" : ""
```
masone, line 55: I would ask the designer to opt for a consistent color instead of introducing switches like this. it's pretty likely they did not notice that they're introducing another variant

src/components/collapsible.tsx
```diff
@@ -30,6 +30,10 @@ interface Props {
   forceCollapse?: boolean
   /** Makes sure that the content is always in the DOM */
   alwaysRender?: boolean
+  /** Teal color chevron */
+  tealChevron?: boolean
+  /** Chevron position on the top for mobile and center for desktop */
+  mobileChevron?: boolean
```
masone, line 36: this sounds like micro-optimization for me and I would try getting around adding boolean props like this. I'd rather opt for a consistent behavior of the collapsible component instead of accommodating for subtle differences in where they are used

masone (changes requested):
can we have a look at this together? I don't understand it in its full detail yet, but it feels like we might be over-optimizing for something the designer would be willing to compromise on

acapop77:
> can we have a look at this together? I don't understand it in its full detail yet, but it feels like we might be over-optimizing for something the designer would be willing to compromise on
Yes sure, I'm available for that. I've done it as visual parts of the collapsible row table component for leads center, after talking to Fred and him pinging back with @lkappeler. This collapsible that we use for row in table are bit different from regular one. Is all here to see: https://github.com/carforyou/carforyou-dealerhub-web/pull/1094

lkappeler:
Probably there was a bit of a misunderstanding between me and Fred...
Basically my intention was to tell him if Aleksander could check on the implementation details on weather the changes he wants / current state comes form the table component and could be handled in there or if its a behavior of the collapsible component as it is apparently.
As for this case to create a ticket and bring it to a refinement so it could be discussed on a proper solution.

acapop77:
> Probably there was a bit of a misunderstanding between me and Fred...
> Basically my intention was to tell him if Aleksander could check on the implementation details on weather the changes he wants / current state comes form the table component and could be handled in there or if its a behavior of the collapsible component as it is apparently.
Yes, chevron colour and position for mobile do. And the 16px on the right hand side we can do that on DH side.
> As for this case to create a ticket and bring it to a refinement so it could be discussed on a proper solution.
I suppose I rushed it - didn't figure it would need a ticket.

masone:
no worries. how will we proceed with this? I interpret we're stepping back from this PR?

acapop77:
> no worries. how will we proceed with this? I interpret we're stepping back from this PR?

Sure, I'll close this PR and the one on DH. Will also make new one on DH just to ajust spacing - one thing we can do on DH side alone. And will get back to this if and when we have proper ticket for it.

PR #319, opened 2021-03-17
feat: menu components

Averethel:
Extracts `MenuItem` and `MenuSectionTitle` components introduced in carforyou/carforyou-header-footer-pkg#112 so that we can use the same basis to create sidebar menu in dealer-hub.

### DIFF
--- a/src/components/menu/item.tsx
+++ b/src/components/menu/item.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentType, FC } from "react"
+import React, { ComponentType, FC, Fragment } from "react"
 import classNames from "classnames"
 
 interface Props {
@@ -22,6 +22,10 @@ interface Props {
    * marks the item as active in the UI
    */
   active?: boolean
+  /**
+   * component to wrap the link with
+   */
+  LinkWrapper?: ComponentType<{ href?: string }>
 }
 
 const MenuItem: FC<Props> = ({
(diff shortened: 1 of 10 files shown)


src/components/menu/item.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { ComponentType, FC } from "react"
+
+interface Props {
+  /**
+   * title to be displayed
+   */
+  text: string
```
masone, line 7: the other has `title`, this one has `text`. worth harmonizing?

src/tailwind/defaultConfig.ts
```diff
@@ -284,6 +284,7 @@ export default {
     letterSpacing: {
       normal: "0",
       wide: "1px",
+      wider: "2px",
```
masone, line 287: next up: widest!

masone (approved):

src/tailwind/defaultConfig.ts
Averethel: haha, that could be and after that we end up with `extra-turbo-mega-widest` :D

In all seriousness though - maybe we should revert to `2xs`-`xs`-`s`-`md`-lg`-`xl`-2xl` scale, which seems to be tailwind default. Fred had been asking if there's a way to change the spacing scale to `8`-`16`-`24`-`32`-`40`-`48`and that's going to be a huge pain if we keep using pixel values in the names...

PR #321, opened 2021-03-24
feat: error page

talamcol:
Reference: [CAR-5386](https://autoricardo.atlassian.net/browse/CAR-5386)

To align the error handling with the listings project in dealerhub, the markup for the error page can be shared. Therefore I added this new component here.

image


src/components/customErrorPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+   */
+  statusCode: number | "default"
+  /**
+   * the source path to the image you want to show
+   */
+  image: string
+  /**
+   * the page title as an h1 heading
+   */
+  pageTitle: string
```
masone, line 17:
```suggestion
  title: string
```

src/components/customErrorPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * the page subtitle as an h2 heading
+   */
+  pageSubtitle: string
```
masone, line 21:
```suggestion
  subtitle: string
```

src/components/customErrorPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * the redirection path on click of the button
+   */
+  redirectHref: string
+  /**
+   * the label for the redirection button
+   */
+  redirectLabel: string
```
masone, line 29: it's not really a redirect, is it?

```suggestion
  buttonLabel: string
```

src/components/customErrorPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+   */
+  pageTitle: string
+  /**
+   * the page subtitle as an h2 heading
+   */
+  pageSubtitle: string
+  /**
+   * the redirection path on click of the button
+   */
+  redirectHref: string
```
masone, line 25:
```suggestion
  buttonUrl: string
```

src/stories/customErrorPage.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { StoryProps } from "./storyProps"
+import StoryContainer from "./storyContainer"
+import CustomErrorPage, { Error404Props } from "../components/customErrorPage"
+
+export default {
+  title: "Error page",
+  component: CustomErrorPage,
+  args: {
+    statusCode: 404,
+    image: "error404.png",
```
masone, line 12: while this might work in the context of storybook, I don't think it will in the project. we'd have to force the image through a require so it's picked up by webpack, no?

src/components/customErrorPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * the label for the redirection button
+   */
+  redirectLabel: string
+}
+
+const CustomErrorPage: FC<Props> = ({
+  statusCode,
+  image,
```
masone, line 34: I would rather encapsulate the image. if we pass everything, there is little value in extracting in the first place

src/components/customErrorPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+  pageTitle,
+  pageSubtitle,
+  redirectHref,
```
masone, line 37: can this just be hardcoded to `/`? what are the paths we're using in the apps?

src/tailwind/defaultConfig.ts
```diff
@@ -506,6 +507,8 @@ export default {
       modalLarge: "770px",
       checkbox: "24px",
       radioIndicator: "12px",
+      110: "110px",
+      200: "200px",
```
ufkaya, line 511: if the values are used only for 1 component then use the CSS module

src/components/customErrorPage.tsx
talamcol: both apps are doing `/${language}`

src/stories/customErrorPage.stories.tsx
talamcol: yes. you have to write `require("~/assets/error404.png")`

should I pass the image as a function?

masone: oh, got it. no worries then. it will check out in the project as the require essentially also returns the resulting path as a string

src/components/customErrorPage.tsx
talamcol: I can also pass it as a function if that is better?

src/components/customErrorPage.tsx
talamcol: but you are right, it works without the language :) it keeps the value

src/tailwind/defaultConfig.ts
talamcol: done

ufkaya (approved):
👍

PR #324, opened 2021-04-06
CAR-6339: feat: extended checkboxFilter component with tracking

GutierrezGo:
While working on the checkboxFilter component extraction, we were skipping to pass the tracking. 
Now, the tracking is added as an optional prop.

PR in listings -> https://github.com/carforyou/carforyou-listings-web/pull/3066

## Thank you 🦙

### DIFF
--- a/src/components/filters/checkbox.tsx
+++ b/src/components/filters/checkbox.tsx
@@ -16,6 +16,7 @@ interface Props {
   apply: (filtersToApply) => void
   selected?: (string | number)[]
   facet?: { [key: string]: number }
+  tracking?: (value) => void
 }
 
 const CheckboxFilter: FC<Props> = ({
@@ -24,11 +25,14 @@ const CheckboxFilter: FC<Props> = ({
   apply: applyFilters,
   selected = [],
   facet = {},
+  tracking,
 }) => {
   const values = options.map(({ value }) => value)
 
   const selectValue = (value) => () => {
     const selectionCopy = [...selected]
+
+    tracking ? tracking(value) : null
 
     if (selectionCopy.includes(value)) {
       const i = selectionCopy.indexOf(value)
--- a/src/components/filters/checkbox.tsx
+++ b/src/components/filters/checkbox.tsx
@@ -16,7 +16,7 @@ interface Props {
   apply: (filtersToApply) => void
   selected?: (string | number)[]
   facet?: { [key: string]: number }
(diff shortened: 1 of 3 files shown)


src/components/filters/checkbox.tsx
```diff
@@ -16,6 +16,7 @@ interface Props {
   apply: (filtersToApply) => void
   selected?: (string | number)[]
   facet?: { [key: string]: number }
+  tracking?: (value) => void
```
Averethel, line 19: Would it make sense to call it `onSelect` or something like that? In the end, it's a function that you call when a value is selected. It's not really limited to tracking.

src/components/filters/checkbox.tsx
```diff
@@ -24,12 +25,15 @@ const CheckboxFilter: FC<Props> = ({
...
   selected = [],
   facet = {},
+  tracking,
 }) => {
   const values = options.map(({ value }) => value)
 
   const selectValue = (value) => () => {
     const selectionCopy = [...selected]
 
+    tracking ? tracking(value) : null
```
Averethel, line 35:
```suggestion
    if (tracking) {
      tracking(value)
    }
```

src/components/filters/checkbox.tsx
Averethel, line 19: Can we add a type to the value here?

Averethel (changes requested):
A good starting point. Please check my comment on naming.

What's missing for me is a test and a story in storybook.

src/components/filters/checkbox.tsx
GutierrezGo: Yes, it makes sense thanks.

GutierrezGo: @Averethel  We can but I was thinking now that this function is not only for tracking maybe is not limited to a type.

src/stories/filters/checkbox.stories.tsx
```diff
@@ -101,3 +101,9 @@ WithIcon.args = {
     },
   ],
 }
+
+export const WithOnSelect = Template.bind({})
+WithOnSelect.args = {
+  storyName: "With onSelect",
+  onSelect: action("on Click"),
```
Averethel, line 108:
```suggestion
  onSelect: action("on select"),
```

Averethel (approved):

PR #326, opened 2021-04-08
feat: extract rangeInput and rangeSelect

Marine-Berthier:
[Ticket 4405 ](https://autoricardo.atlassian.net/browse/CAR-4405?atlOrigin=eyJpIjoiNmQ1OGNjNjRkMTIwNGNjNTg4NTgxMGY4YWVmODdmZDkiLCJwIjoiaiJ9
)
Extract the range selects in an input- and select-variation to the components library package. Replace the usage in the listings and the dealerhub projects.

Related to [this PR](https://github.com/carforyou/carforyou-dealerhub-web/pull/1187) and [this one ](https://github.com/carforyou/carforyou-listings-web/pull/3086)


src/components/fieldHelpers/label.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { FC } from "react"
+
+interface Props {
+  label?: string
+  required?: boolean
+  htmlFor?: string
```
Averethel, line 6: Can we call it `fieldName` or something more descriptive?

src/components/fieldHelpers/label.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const Required: FC = () => (
+  <span className="text-salmon leading-xxs relative top-requiredIndicator">
+    &nbsp;*
+  </span>
+)
+
+const Label: FC<Props> = ({ label, required = false, htmlFor }) => {
+  if (!label) return null
+  htmlFor = htmlFor || label
```
Averethel, line 17: What's the motivation behind this? Wouldn't this result in a translated string being set as `for` attribute?

src/components/fieldHelpers/label.tsx
Averethel, line 6: Instead of using `for` we can also wrap the input in the label tag:

```html
<label>
   I am label
   <input name="test" type="text" />
</label>
```
Maybe this approach would simplify the interface?

src/components/rangeInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import Input from "./input/index"
+
+import Label from "./fieldHelpers/label"
+
+interface Props {
+  name: {
+    min: string
+    max: string
+  }
```
Averethel, line 11: Couldn't we pass a single name and construct `min` and `max` names with an appropriate suffix?

src/components/rangeInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+    min: boolean
+    max: boolean
+  }
+  placeholder?: {
+    min: string
+    max: string
+  }
+  required?: boolean
+  rangeInputLabel?: string
+  initialValue?: string
```
Averethel, line 28: What's the difference between `initialValue` and `value`? Why is this one a string?

Averethel, line 28: Oh, it's not used in the component. Can you remove it?

src/components/rangeInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import Label from "./fieldHelpers/label"
+
+interface Props {
+  name: {
+    min: string
+    max: string
+  }
+  handleChangeMin: (value) => void
+  handleChangeMax: (value) => void
```
Averethel, line 13: We pass other props as objects. Can we align this:

```suggestion
  handleChange: {
    min: (value) => void
    max: (value) => void
  }
```

src/components/rangeSelect.tsx
```diff
@@ -0,0 +1,69 @@
...
+  isValid?: {
+    min: boolean
+    max: boolean
+  }
+  placeholder?: {
+    min: string
+    max: string
+  }
+  required?: boolean
+  options?: Array<{ name: string; value: number | { customValue: number } }>
```
Averethel, line 25: Would range select ever accept a custom value? I think we can just roll with:

```suggestion
  options?: Array<{ name: string; value: number }>
```

src/components/rangeSelect.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import Select from "./select"
+
+interface Props {
+  name: {
+    min: string
+    max: string
+  }
+  handleChangeMin: (value) => void
+  handleChangeMax: (value) => void
```
Averethel, line 11: We pass other props as objects. Can we align this:

```suggestion
  handleChange: {
    min: (value) => void
    max: (value) => void
  }
```

Averethel (changes requested):
A very good starting point, nice job.
I have a couple of questions and suggestions. What is missing for me are some tests too.

src/components/rangeInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+  isValid?: {
+    min: boolean
+    max: boolean
+  }
+  placeholder?: {
+    min: string
+    max: string
+  }
+  required?: boolean
+  rangeInputLabel?: string
```
Averethel, line 27: Any reason to not just call it label?

src/components/rangeSelect.tsx
```diff
@@ -0,0 +1,69 @@
...
+    min: boolean
+    max: boolean
+  }
+  placeholder?: {
+    min: string
+    max: string
+  }
+  required?: boolean
+  options?: Array<{ name: string; value: number | { customValue: number } }>
+}
```
Averethel, line 26: Would select not need a label too?

src/components/fieldHelpers/label.tsx
Marine-Berthier: Yes definitely !

src/components/rangeInput.tsx
Averethel, line 13: I noticed in related project PRs that you're passing the same change handler. Is the separation here really worth it then?

src/components/rangeInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <Label
+          label={rangeInputLabel}
+          htmlFor={`${minName}-min`}
+          required={required}
+        />
+      ) : null}
+      <div className="flex flex-wrap">
+        <div className="w-6/12">
+          <Input
+            name={minName}
```
talamcol, line 52: The input field already supports label. Can you please check if you could leverage it here too?
Something like:
```jsx
<Input
            labelText={rangeInputLabel ? rangeInputLabel : null}
            required={required}
```
on the first input field.
Then you may drop the new label component

src/components/rangeSelect.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+  isValid?: {
+    min: boolean
+    max: boolean
+  }
+  placeholder?: {
+    min: string
+    max: string
+  }
+  required?: boolean
```
talamcol, line 24: It looks like you are not passing this prop down to the component

src/components/rangeSelect.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { FC } from "react"
+
+import Select from "./select"
+
+interface Props {
+  name: {
```
talamcol, line 6: I'm not sure if this would be nicer but since you have that many props for min/max maybe we could have 1 object defining everything for min and another one defining everything for max. What do you think?

src/components/rangeInput.tsx
talamcol, line 27: I would rename this to label only

src/components/rangeInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const RangeInput: FC<Props> = ({
+  name: { min: minName, max: maxName },
+  handleChangeMin,
+  handleChangeMax,
+  value: { min: minValue = "", max: maxValue = "" },
+  placeholder: { min: minPlaceholder = "", max: maxPlaceholder = "" },
+  rangeInputLabel,
+  required,
```
talamcol, line 38: could you please define a default for this optional prop? Thanks!

src/components/fieldHelpers/label.tsx
Marine-Berthier: I don't think we need it actually. I used a label component from DH and did not clean enough !

src/components/rangeInput.tsx
Marine-Berthier: Yes 👍

src/components/rangeInput.tsx
Marine-Berthier: Yes, not needed (for now)

src/components/rangeSelect.tsx
Marine-Berthier: 👍 Indeed, In that case we would go for the input version I guess !

src/components/rangeInput.tsx
Marine-Berthier: Yes because Storybook complains : "label" is already one of its args

src/components/rangeSelect.tsx
Marine-Berthier: From my understanding, this component is used only in listings as a filter and is wrapped within 
`<CollapsibleFilter
      title={t("listing.seats.label")}
      isCollapsed={true}
      mobileOnly={mobileOnly}
      name="seats"
      key="seats"
    >` which returns the label. 
    Writing this, I realised that we did not really need to extract this component since it is a not a shared one. What do you think ?

src/components/rangeInput.tsx
Marine-Berthier: No, we don't need both !

src/components/rangeSelect.tsx
Marine-Berthier: Yes good point !

src/components/rangeInput.tsx
Marine-Berthier: I wanted to do that at first as well but Storybook has already a props called "label". I will double check though !

src/components/rangeSelect.tsx
Marine-Berthier: Or as @Averethel mentioned, define only one name, value and so on and use suffix for min and max

src/components/fieldHelpers/label.tsx
Marine-Berthier: I don't know because for me, we need to be able to choose to add a label or not, depending whether the component is used as a filter or not.

src/components/rangeInput.tsx
Averethel: Can we try to solve it in storybook? Right now we make our external interface more complex because the internal documentation tool is not happy with the naming....

src/components/rangeInput.tsx
Marine-Berthier: I tried your suggestion and it does not work because we have 2 inputs and they are not on the same level anymore. I guess I could make it work with some css, but it sounds a bit shaky

src/components/rangeSelect.tsx
Averethel: I'd say let's not extract it at the moment then. We should've split the ticket per component anyway.

src/components/rangeInput.tsx
Marine-Berthier: Yes, I fixed it and called it `label` 👍

assets/input/basic.css
```diff
@@ -87,7 +87,8 @@ textarea[data-valid="false"]:focus {
   @apply text-grey-dark;
 }
 
-input.input_left {
+input.input_left,
+div.input_left {
```
Averethel, line 91: Do we need a `div` tag here? We use the `input` tag to ensure specificity and catch all the browser overrides and such.

src/components/__tests__/rangeInput.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+      }}
+      handleChange={handleChange}
+    />
+  )
+
+describe("RangeInput", () => {
+  it("renders RangeInput correctly", () => {
+    const { container } = renderWrapper()
+    expect(container).toMatchSnapshot()
+  })
```
Averethel, line 35: I'm not sure this test brings a lot of value. The resulting markup is large and the slightest change will result in this test failing. How about checking that there are two input elements with expected names instead?

src/components/__tests__/rangeSelect.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+      }}
+      options={options}
+    />
+  )
+
+describe("RangeSelect", () => {
+  it("renders RangeSelect correctly", () => {
+    const { container } = renderWrapper()
+    expect(container).toMatchSnapshot()
+  })
```
Averethel, line 49: I'm not sure this test brings a lot of value. The resulting markup is large and the slightest change will result in this test failing. How about checking that there are two selects with expected names instead?

Averethel (approved):

assets/input/basic.css
Marine-Berthier: I should have used input you're right ! I am changing it

PR #327, opened 2021-04-08
fix: menu item key

talamcol:
If the key is not defined on the options, it will use the value as the key. Because the value is often an object such as `{sortOrder: "ASC", sortType: "PRICE"}`, `value.toString()` will result in `[object Object]`.

As a result, the DOM is invalid because of duplicated keys

image

### DIFF
--- a/src/components/dropdown/menu.tsx
+++ b/src/components/dropdown/menu.tsx
@@ -127,17 +127,15 @@ class Menu<T> extends Component<Props<T>> {
             padding
           )
 
-          const key =
-            item.key || typeof item.value === "object"
-              ? `item-${index}`
-              : item.value?.toString()
-
           return (
             // eslint-disable-next-line react/jsx-key
             <li
               data-testid={item.name}
               {...getItemProps({
-                key,
+                key:
+                  item.key ||
+                  (typeof item.value !== "object" && item.value.toString()) ||
+                  `item-${index}`,
                 item,
                 className: classNames(
                   "hover:bg-grey-bright transition duration-200 cursor-pointer",
--- a/src/components/dropdown/menu.tsx
+++ b/src/components/dropdown/menu.tsx
@@ -127,12 +127,17 @@ class Menu<T> extends Component<Props<T>> {
             padding
           )
 
+          const key =
+            item.key || typeof item.value === "object"
+              ? `item-${index}`
+              : item.value?.toString()
+
           return (
             // eslint-disable-next-line react/jsx-key
             <li
               data-testid={item.name}
               {...getItemProps({
-                key: item.key || item.value?.toString() || `item-${index}`,
+                key,
                 item,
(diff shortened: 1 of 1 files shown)


src/components/dropdown/menu.tsx
```diff
@@ -127,12 +127,17 @@ class Menu<T> extends Component<Props<T>> {
             padding
           )
 
+          const key =
+            item.key || typeof item.value === "object"
+              ? `item-${index}`
+              : item.value?.toString()
```
Averethel, line 133: The ternary isn't equivalent to what we did before. Consider the case of a `null` value. Since the type of null is an object it would result of `undefined` being returned as key which would result in another error.
Wouldn't this do the trick:
```suggestion
          const key = item.key || (item.value !== "object" && item.value?.toString()) || `item-${index}`
```

Averethel (changes requested):

src/components/dropdown/menu.tsx
talamcol: You are right, sorry. Close, you missed the `typeof`

item.key || (typeof item.value !== "object" && item.value.toString()) || `item-${index}`

Thanks for the help!

Averethel (approved):

src/components/dropdown/menu.tsx
```diff
@@ -132,7 +132,10 @@ class Menu<T> extends Component<Props<T>> {
             <li
               data-testid={item.name}
               {...getItemProps({
-                key: item.key || item.value?.toString() || `item-${index}`,
+                key:
+                  item.key ||
+                  (typeof item.value !== "object" && item.value.toString()) ||
+                  `item-${index}`,
```
masone, line 138: the last option seems to always work. do we really need a fallback chain of 3 options?

talamcol: are you suggesting to always use `item-${index}` ?

masone: yes. we just need a unique key per item and that seems to do the trick? is there anything I'm missing? what if multiple of those are on the page?

talamcol: yes. basically it's only a react thing. Multiple keys that are the same are throwing a warning. That is what's happening at the moment

talamcol: https://reactjs.org/docs/lists-and-keys.html#keys

you should not use the index as a key if the order of the items change and this might negatively impact the performance

masone (approved):
your call! my question should not prevent you from merging. it was like this before :)

src/components/dropdown/menu.tsx
talamcol: As discussed in Slack, since we are using the `<Menu />` component also for the selects with dynamic data, I'll keep the approach as it is

PR #328, opened 2021-04-09
fix: fixed applyFilter function

GutierrezGo:
References https://autoricardo.atlassian.net/browse/CAR-7009

While working on the inputFilter extraction, I found that when we apply the value on the type input in the Make & model section, the last character is deleted. This issue affects already DH.

Listings PR -> https://github.com/carforyou/carforyou-listings-web/pull/3073
DH PR -> https://github.com/carforyou/carforyou-dealerhub-web/pull/1191

## Thank you 🐞

### DIFF
--- a/src/components/filters/input.tsx
+++ b/src/components/filters/input.tsx
@@ -14,7 +14,7 @@ interface Props {
   min?: number
   max?: number
   position?: "left" | "right"
-  apply(event)
+  apply: (event) => void
 }
 
 interface State {
--- a/src/components/filters/input.tsx
+++ b/src/components/filters/input.tsx
@@ -14,7 +14,7 @@ interface Props {
   min?: number
   max?: number
   position?: "left" | "right"
-  apply: (filtersToApply) => void
+  apply(event)
 }
 
 interface State {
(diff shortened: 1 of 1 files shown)


src/components/filters/input.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
   min?: number
   max?: number
   position?: "left" | "right"
-  apply: (filtersToApply) => void
+  apply(event)
```
ufkaya, line 17: you are not defining a type here, should be `(event) => void`

ufkaya (approved):

src/components/filters/input.tsx
```diff
@@ -14,7 +14,7 @@ interface Props {
   min?: number
   max?: number
   position?: "left" | "right"
-  apply: (filtersToApply) => void
+  apply: (event) => void
```
Averethel, line 17: I'm a bit late to the party but is there no way around using raw event here?
Other filters use the object interface and it would be good to have consistency...
Also, keep in mind that changes like this are breaking changes

masone: also late to the party. this was a breaking change :)

PR #330, opened 2021-04-13
fix: change <a/> to <div/> to allow links within the collapsible

talamcol:
Reference: [CAR-7124](https://autoricardo.atlassian.net/browse/CAR-7124)

In some pages in DH and listings, there is an error in the console that the DOM is invalid. The reason for it is that the collapsible itself is an `<a/>` tag. This does not allow to have other a tags within the collapsible component. As we have this requirement, I changed it here to a div.
image

### DIFF
--- a/src/components/collapsible.tsx
+++ b/src/components/collapsible.tsx
@@ -48,9 +48,9 @@ const Collapsible: FC<Props> = ({
 
   return (
     <>
-      <a
+      <div
         className={classNames("flex items-center cursor-pointer", {
-          "hover:opacity-100": !opacityOnHover,
+          "transition duration-200 hover:opacity-60": opacityOnHover,
         })}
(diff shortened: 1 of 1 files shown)


Averethel (commented):
What would be the case where we have a link as a collapsible toggle?
Does that even make sense? We would navigate instead of opening colapsible, no?

talamcol:
> What would be the case where we have a link as a collapsible toggle?
> Does that even make sense? We would navigate instead of opening colapsible, no?

I think so yet. We would need to prevent the default if we don't want that. But I think passing an <a/> as the toggle also does not make sense from an SEO perspective when it's not an actual link, no? 
I quickly checked listings and DH and we never pass a as a toggle

Averethel (approved):

PR #335, opened 2021-04-23
fix: do not pass href prop to fragment

Averethel:
In case the menu item doesn't have a link wrapper we pass `href` prop to `React.Fragment`.
This results in warnings from react:
Screenshot 2021-04-23 at 10 08 50

## After
`href` prop is only passed when we don't use `Fragment` as default wrapper

### DIFF
--- a/src/components/menu/item.tsx
+++ b/src/components/menu/item.tsx
@@ -25,7 +25,7 @@ interface Props {
   /**
    * component to wrap the link with
    */
-  LinkWrapper?: ComponentType<{ href: string }>
+  LinkWrapper?: ComponentType<{ href?: string }>
 }
 
 const MenuItem: FC<Props> = ({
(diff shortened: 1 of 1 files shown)


src/components/menu/item.tsx
```diff
@@ -34,10 +34,12 @@ const MenuItem: FC<Props> = ({
...
   url,
   active = false,
-  LinkWrapper = Fragment,
+  LinkWrapper,
 }) => {
+  const LinkWrapperComponent = LinkWrapper || Fragment
+
   return (
-    <LinkWrapper href={url}>
+    <LinkWrapperComponent {...(LinkWrapper ? { href: url } : {})}>
```
ufkaya, line 42: Since URL is optional, shouldn't you check that's valid

ufkaya (approved):

src/components/menu/item.tsx
Averethel: `href` is optional for the wrapper component too, if the `url` is not passed we will pass `{ href: undefined }` which is fine.

PR #339, opened 2021-05-03
feat(input): adds support for input debounce

gzimh:


src/components/input/inputField.tsx
```diff
@@ -61,8 +59,15 @@ interface Props extends InputHTMLAttributes<HTMLInputElement> {
   hasError?: string
   disabled?: boolean
   required?: boolean
-  autoComplete?: string
-  onChange?: (e: ChangeEvent<HTMLInputElement>) => void
+  onChange?: (
+    e:
+      | ChangeEvent<HTMLInputElement>
+      | {
```
lkappeler, line 65: This looks a bit like a workaround to me, what's the reason why we need to handle a custom interface and cant we update the source so where we call change with a different event instead to avoid having to support a custom interface?

src/components/input/inputField.tsx
```diff
@@ -78,7 +83,6 @@ const InputField = forwardRef<HTMLInputElement, Props>(
       hasError = false,
       disabled = false,
       required = false,
-      autoComplete,
```
lkappeler, line 81: How does dropping the` autoComplete` prop relate to the denounce change?

src/components/input/index.tsx
```diff
@@ -99,15 +101,18 @@ const Input = forwardRef<HTMLInputElement, Props>(
       visible: !!value,
       disabled,
       onClear: () => {
-        onChange({ target: { name, value: "", cleared: true } })
+        const clearEvent = { target: { name, value: "", cleared: true } }
+        onChange(clearEvent)
+        setCurrentEvent(clearEvent)
       },
     }
+    const [currentValue, setCurrentEvent] = useDebounce(onChange, debounce)
```
lkappeler, line 109: Can you make them consistent so the setter and value have the same name.

lkappeler (commented):
Can you describe what was off with the current implementation and your approach to fixing it, I think it looks fine but it's quite hard to judge looking at the diff of the PR...

Also, I think we should replace the tests with some targeting the new feature.

Not fully sure if I got that correctly from the change-set, feel free to correct me where I got something wrong...
So, in general, our debounce works in delaying the call of the passed onChange callback by the given amount of ms?
- What I wondered is if we have an issue with the changed behaviour regarding the js stack, before if there was no delay or a delay of 0 we would not wrap the callback execution in a setTimeout call so in that way, will execute the CB directly whereas wrapping it in a setTimeout of 0ms will push it outside the stack and execute it after all pending synchronous operations are complete...

src/components/input/inputField.tsx
gzimh: Oops, this was a mistake! Nothing to do with debounce

masone:
If you two get into a call to discuss this, I'd love to understand this better as well. There are some issues I remember that might be vaguely related

Averethel:
@gzimh @lkappeler what's the staus here? Can we move forward?

lkappeler:
@Averethel I need to find some time to wrap my head around, hopefully, tomorrow if you have capacity feel free to take over :) just let me know.

Averethel:
@lkappeler all good thanks for taking care 👍🏻

lkappeler:
So I had some time to look into the topic...
In general, I think it goes in the right direction with what you wanna achieve with updating the input state and calling the callback after the denounce.

Regarding the `event.persist()` I am a bit worried that this is a rather short term solution as they say with v17 event pooling will be dropped...

I looked a bit in the usage of our input callbacks and I think we can update it towards debouncing values or like a defined object as the primary use case is to update something based on the new value (found t[his artcle](https://medium.com/@gabrielmickey28/using-debounce-with-react-components-f988c28f52c1#9d21)) It will allow us to avoid the problems with the event target that is gone while still providing the needed values to update the app state. something like 
```
{ name, cleared, value } = event.target
```
the `cleared`, we might be able to factor out 

Before you start I would like to get @Averethel or @masone opinion on the value-based debounce approach so don't head in the wrong direction. (Also from my perspective at this point consider updating the bug to a tech task) 

Please DM me if you have questions or feel like it's simpler to have a chat on it

disufi:
@lkappeler Could we try to go with this solution and look into the alternative (react 17 compatible) approach in a follow-up? What do you think?

lkappeler:
From my point of view, I would avoid implementing a fix that will create more work when we start tackling the update for the following reasons:

- The functionality is broken atm and it has been worked around in the components package and the concrete projects. We kinda already created a follow up with this bug to properly tackle the case when implementing the feature the bug was created from. By going with an incompatible approach with the current react version I fear that we are creating an overhead of work by creating an in-between version that will increase the refactoring efforts when tackling the update.
- The projects are already using custom debounce implementations in wrappers etc (components lib as [well](https://github.com/carforyou/carforyou-components-pkg/blob/master/src/components/filters/input.tsx#L50) that probably need to be updated towards using the functionality from the base input so I would rather define a future proof approach
- We are more than 6 months behind with updating react so it is something we wanna do in the future for sure and I don't wanna create more effort than already needed

What I would do is break the interface and return the new value as a string instead of an event, along with keeping the interface for the delay and replace the usage across the projects as a fix and form there create follow-ups to refactor out the custom implementations withing the components lib and the consuming projects so we can get to a state where we rely on the implementation in a single place to keep it maintainable.

That being said I don't have a strong opinion on how to organize it in the form of tickets and sprint planning, we should stay on the topic and find a way forward and keep an eye on not creating overhead by seeking a fast way to close the task.

masone:
I'm with Lorenzo here. however, feel free to approach it your preferred way if you commit to tackling it immediately. We can split tickets and PRs sensibly, but I'd like to follow through on Lorenzos proposal in coming days and weeks. As you'll be the one working on it, feel free to approach it the way you think brings you to success

lkappeler:
Had a discussion with Gzim:

He pointed out that he expects the implementation to work fine with react 17 as their breaking change was to remove a backwards compatibility implementation on top of the browsers events.

https://reactjs.org/docs/legacy-event-pooling.html
> This is because React reused the event objects between different events for performance in old browsers, and set all event fields to null in between them. With React 16 and earlier, you have to call e.persist() to properly use the event, or read the property you need earlier.

>In React 17, this code works as you would expect. The old event pooling optimization has been fully removed, so you can read the event fields whenever you need them.

> This is a behavior change, which is why we’re marking it as breaking, but in practice we haven’t seen it break anything at Facebook. (Maybe it even fixed a few bugs!) Note that e.persist() is still available on the React event object, but now it doesn’t do anything

We are not fully sure if I will just work or the value will be empty. He will create a PoC (locally with build react app with react 17) to see if it works or not.

If it works we tackle the unit test and go with the implementation otherwise we will discuss how to continue.

gzimh:
I've created a PoC with react: `^17.0.2` and I can confirm that this solution works well 🎉
As the docs mention, calling `event.persist()` in react 17 doesn't have any effect. But we need to explicitlly call it on 16.x.x.

src/components/input/index.tsx
```diff
@@ -124,7 +129,15 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
         required={required}
-        onChange={onChange}
+        onChange={(e: ChangeEvent<HTMLInputElement>) => {
+          e.persist()
+          setInputEvent({
+            ...e,
+            target: {
+              ...e.target,
+            },
+          })
```
gzimh, line 139: This may look weird but it was the only workaround to make the unit test pass. For some reason, if I pass the `e` directly, the `e.target.value` was passed as empty when in reality it contains value. This caused the test to fail.
This only happend in testing environment.

src/components/input/index.tsx
```diff
@@ -99,15 +101,18 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
       },
     }
+    const [inputValue, setInputEvent] = useDebounce(onChange, debounce)
 
     const renderInput = (hasError) => (
       <InputField
         ref={ref}
         name={name}
-        value={value || ""}
+        value={inputValue}
```
masone, line 115: I interpret that `inputValue` would be the debounced one. is that really what we want as the value of the input field?

src/components/input/index.tsx
```diff
@@ -124,7 +129,10 @@ const Input = forwardRef<HTMLInputElement, Props>(
         hasError={hasError}
         disabled={disabled}
         required={required}
-        onChange={onChange}
+        onChange={(e: ChangeEvent<HTMLInputElement>) => {
+          e.persist()
```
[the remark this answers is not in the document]
masone: we're still relying on persisting the event. if I remember correctly, Lorenzo pointed out that this is deprecated in the current React version. can you elaborate on the plan again?

src/components/input/index.tsx
masone: this sounds _very_ odd. I'd prefer to understand what is going on exactly before merging with this workaround

masone:
What actions will we need to plan to bring this change to the projects? Do you plan to do them with your current task or would we create additional ones?

src/components/input/index.tsx
gzimh: No, actually `inputValue` is the immidiate value, the debonced value is passed to the callback function (useDebounce.tsx)

gzimh:
>we're still relying on persisting the event. if I remember correctly, Lorenzo pointed out that this is deprecated in the current React version. can you elaborate on the plan again?

We are clear in that regard, react 17 wont have any breaking changes. Basically they will implement the `e.presist()` behaviour by default and calling `e.presist()` will make no difference. But in the current version it's a must that we call it.

gzimh:
> this sounds very odd. I'd prefer to understand what is going on exactly before merging with this workaround
Yes indeed. It's needed for tests only. In real-world executions, the `e` is passed without losing the `target.value` as is expected. 
I assume something is going on with tests that causes the `e.target.value` to be empty, so, as a workaround I copied the object state to a new object and passed that instead.

gzimh:
>What actions will we need to plan to bring this change to the projects? Do you plan to do them with your current task or would we create additional ones?

https://github.com/carforyou/carforyou-dealerhub-web/pull/1326 I did only one for now, its for the search input on Contact Requests (DH)

Basically all we would need to do is to pass in the debounce prop to Input component

src/components/input/index.tsx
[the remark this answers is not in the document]
masone: addressed in the PR comments

lkappeler:
Update: Colin added  the changes discussed with Pascal in a stacked PR: https://github.com/carforyou/carforyou-listings-web/pull/3257

gzimh:
After another discussion with @lkappeler  we decided that the workaround (object copying) is not worth it just so the test can pass. So, we decided to skip the test completely for now.

lkappeler (approved):

src/lib/__tests__/debounceHelper.test.ts
```diff
@@ -1,69 +0,0 @@
-import debounce from "../debounceHelper"
-
-describe("debounceHelper", () => {
```
masone, line 3: you said "skip the test", which I would be ok with if we commit to unskipping it asap. I'm not so much ok with dropping all of them, which removes our accountability for readding it completely

gzimh: @masone okay as i can see we have two options:
* merge this without the tests and create a ticket for the next sprint to add the tests.
* leave this as spillover for the next sprint and merge only after adding tests

Which one do u prefer?

PR #342, opened 2021-05-04
fix: change padding

ufkaya:
References [CAR-7194](https://autoricardo.atlassian.net/browse/CAR-7194)

## Motivation and context

On iPhone 7 and iPhone X, the Calendar for test drive request are more towards the right of the screen.

## How to test
Go to https://car7194.carforyou-listings-web.branch.carforyou.ch/
Check that the calendar is not wider than the input

### DIFF
--- a/src/components/datePicker/index.module.css
+++ b/src/components/datePicker/index.module.css
@@ -12,12 +12,6 @@
     @apply shadow-hard rounded-6 bg-white leading-xs w-12/12 p-10;
 }
 
-@screen xxs {
-    :local .react-calendar {
-        min-width: 346px;
-    }
-}
-
 :local .react-calendar,
 :local .react-calendar *,
 :local .react-calendar *:before,
--- a/src/components/datePicker/index.module.css
+++ b/src/components/datePicker/index.module.css
@@ -12,6 +12,12 @@
     @apply shadow-hard rounded-6 bg-white leading-xs w-12/12 p-10;
 }
 
+@screen xxs {
+  :local .react-calendar {
+      min-width: 346px;
+  }
+}
+
 :local .react-calendar,
 :local .react-calendar *,
 :local .react-calendar *:before,
--- a/src/components/modal/index.tsx
+++ b/src/components/modal/index.tsx
@@ -74,11 +74,11 @@ const Modal: FC<ModalProps> = ({
           <CloseMIcon color={style === "dark" ? "#FFFFFF" : "#232A36"} />
         </div>
         <div
-          className={classNames("z-modal w-12/12 my-0", {
-            "p-40 md:w-modalLarge": size === "large",
-            "p-40 md:w-modal": size === "medium",
-            "p-40 md:w-modalSmall": size === "small",
-            "h-full": size === "fullscreen",
+          className={classNames("z-modal w-12/12 my-0 p-15", {
+            "md:p-40 md:w-modalLarge": size === "large",
+            "md:p-40 md:w-modal": size === "medium",
+            "md:p-40 md:w-modalSmall": size === "small",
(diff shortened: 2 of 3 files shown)


src/components/datePicker/index.module.css
```diff
@@ -12,12 +12,6 @@
     @apply shadow-hard rounded-6 bg-white leading-xs w-12/12 p-10;
 }
 
-@screen xxs {
-    :local .react-calendar {
-        min-width: 346px;
```
ufkaya, line 17: does anyone remember why we need this value?

lkappeler: to ensures the files stay square and don't become rectangular, simply dropping it will not make that sure any more, if it does not has the required width we probably need to lower font and tile size one breakpoint before...
(Only gave it a quick test there might be more UI related issues when just dropping it)

Before:
Screenshot 2021-05-05 at 17 55 19

After
Screenshot 2021-05-05 at 17 55 41

After:  (font not centered)
Screenshot 2021-05-05 at 17 48 18

lkappeler: Something that just came to my mind, using the calendar as a dropdown option is only one use case for the component. As a component, we might wanna use it for other purposes where the calendar is not spaced. In that way, the component does not push itself to the right, it's the listings project that does that. I am not fully sure about the desired behaviour from a UX perspective in the concrete use case but the component will never become wider than the screen. In that way, if the component ensures its UI works as intended with the min-width and would look perfectly fine when given the available space before becoming smaller...

image

ufkaya: thanks for checking, the issue is that the calendar can be wider than the input element, and might look weird, I will ask the designer if its fine or not, if its not a bug then I will just close this PR

lkappeler: Yes, I saw the ticket. (Not sure if the intended solution is that the calendar should open centre to the input in that case or we should show the smaller version of the calendar) What I tried to explain is that we can not just drop the min within the component as it breaks the intended behaviour to fix one use case.

For example, in the B version of the PDP experiment, the calendar is also used in the non-modal contact form (mad it one unit smaller to show the effect better) by dropping the min-width the calendar gets squeezed to the row and does not show up as in intended anymore.

Without min-width
Screenshot 2021-05-06 at 09 51 12

Current implmentation
Screenshot 2021-05-06 at 09 52 19

I am not entirely sure how I would approach it ( also depends on the desired behaviour of the designer) but I think we can not work around the bug by dropping the min width

ufkaya: I just checked with the designer, we decided to keep the min-width and center the calendar if possible

ufkaya: yeah I didnt know why min-width was needed thats why asked about it

ufkaya:
will fix it on the listings project

src/components/modal/index.tsx
```diff
@@ -74,11 +74,11 @@ const Modal: FC<ModalProps> = ({
           <CloseMIcon color={style === "dark" ? "#FFFFFF" : "#232A36"} />
         </div>
         <div
-          className={classNames("z-modal w-12/12 my-0", {
-            "p-40 md:w-modalLarge": size === "large",
-            "p-40 md:w-modal": size === "medium",
-            "p-40 md:w-modalSmall": size === "small",
-            "h-full": size === "fullscreen",
+          className={classNames("z-modal w-12/12 my-0 p-15", {
```
lkappeler, line 77: Did you verify with the designer that it's fine to change the default padding for every modal on mobile?

ufkaya: Yes, he noticed that the padding is too big on mobile

lkappeler (approved):

PR #346, opened 2021-05-12
feat: add verified ribbon component

mreljin:
Add the verified ribbon as it's used both in listings and DH. 

image


Averethel (approved):
Can you add stories to the storybook so we can showcase the component?

PR #347, opened 2021-05-14
fix: align button style

Averethel:
References https://autoricardo.atlassian.net/browse/CAR-7219

Other than adding a new disabled style for a button with a teal border this PR fixes how we set button style.
Before this was controlled by boolean props `teal` and `tealBorder`, which could lead to setting an "impossible" style (by setting both to true). This has now been replaced by `style` prop which accepts `salmon` (default), `teal` and `teal-border`.

Because of this, it will be a BREAKING CHANGE.


GutierrezGo (approved):
Thank you 🌈

PR #348, opened 2021-05-18
feat: introduce tabs component

Averethel:
Implements CAR-7324

This PR adds a new `Tabs` component:

**Single tab**
Screenshot 2021-05-18 at 15 04 14

**Multiple tabs**
Screenshot 2021-05-18 at 15 04 01

**Tabs with icons**
Screenshot 2021-05-18 at 15 05 00

### DIFF
--- a/src/components/tabs.tsx
+++ b/src/components/tabs.tsx
@@ -2,7 +2,7 @@ import React, { FC, ReactNode, useState } from "react"
 import classnames from "classnames"
 
 interface Tab {
-  name: string
+  title: string
   renderIcon?: () => ReactNode
   renderContent: () => ReactNode
 }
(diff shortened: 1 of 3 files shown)


src/components/tabs.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { FC, ReactNode, useState } from "react"
+import classnames from "classnames"
+
+interface Tab {
+  name: string
```
GutierrezGo, line 5: I'm not so good with names. But we could find another name for this prop, maybe `label`, `text`. 
`name` is too abstract in my point of view. But up to you.

GutierrezGo (approved):
It looks good to me.

src/components/tabs.tsx
Averethel: Makes sense, I went with `title` seems to be more specific and encapsulates what the prop actually does. Thanks for the push.

PR #349, opened 2021-05-18
fix: tabs jumping issues

Averethel:
- due to the border being only rendered when the tab is not active text was jumping by 1px when switching between active/inactive. Since tailwind does not support variable border colours we fix it by adding margin when the tab is active
- the border separating tabs had corner pixel missing since it was only rendered on one side. Negative margin and rendering it on both sides fixes it.


talamcol (approved):
another option would have been to show the border always and just change the color to transparent. But this also looks good 👍

masone (approved):

masone:
I like Colins approach :)

Averethel:
@masone @talamcol that was my initial approach, but tailwind only supports global (not per side) border colours. Since we always have left and right borders to divide the tabs that would require reaching out for a CSS module so I decided against it.

PR #354, opened 2021-05-31
feat: add notification icon

Marine-Berthier:
[CAR-7329](https://autoricardo.atlassian.net/browse/CAR-7329?atlOrigin=eyJpIjoiMGM4MDMzZjgzNTY1NDJhZDliZTFkMDI1NzJhMGU0MjQiLCJwIjoiaiJ9)

Add notification icon // for [this PR ](https://github.com/carforyou/carforyou-dealerhub-web/pull/1303)


src/components/icons/notification.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { FC } from "react"
+
+const Notification: FC = () => (
```
GutierrezGo, line 3: I think you don't need a functional component for this.
Maybe just a function like:
`const Notification = () => (...)`

src/components/menu/item.tsx
```diff
@@ -55,6 +60,11 @@ const MenuItem: FC<Props> = ({
           <IconComponent width="24px" height="24px" />
         </span>
         {title}
+        {NotificationComponent ? (
+          <span className="w-12/12 text-right text-salmon px-5">
```
GutierrezGo, line 64: Not sure about the use of the `text-salmon` style here. Can you elaborate on the use of that style?

GutierrezGo:
In case that the PR is ready for reviews, I will suggest you request FE 👍

src/components/icons/notification.tsx
Marine-Berthier: Yes, I changed it 👍

src/components/menu/item.tsx
Marine-Berthier: Thanks for noticing this Virginia ! At the beginning my notification was black in DH without this class here, but it is no more the case so I removed it

src/components/menu/item.tsx
```diff
@@ -55,6 +60,11 @@ const MenuItem: FC<Props> = ({
           <IconComponent width="24px" height="24px" />
         </span>
         {title}
+        {NotificationComponent ? (
+          <span className="w-12/12 text-right px-5">
+            <NotificationComponent width="14px" height="14px" />
```
masone, line 65: we have a bit of a mix between two patterns here...
if you'd like to have the notification icon in the components-pkg (which I think is a good idea), there is no point in passing a component. a `hasNotification` boolean prop would do the job. that's good to force a specific style, sizing, etc.

in contrast, if you expect the designer to want to customize the notification per project or per use-case, passing the component would be preferred. in that case, you could delete the Notification component in here

masone: oh, I realize you might have added it for the sake of storybook?

masone: I think the two cases are still valid to think about... will you ever want to render any other component there? or can that be the responsibility of the MenuItem that noone outside of here has to worry about?

Marine-Berthier: Thanks for the feedback @masone ! 
Yes, the Notification component is here for the sake of StoryBook. 
I got what you mean though.  I just wanted to leave the choice to return any component in case we want to use this NotificationComponent for something else in the future. For example we could still use it to return this red dot, and in addition, return a different icon for another menuItem. 
Right now there is no value in it and I agree that it would be better not to worry about which component to return outside of MenuItem and just go with a boolean ! 👍

masone: ok let's go for that then! makes using the component simpler in the project too!

masone (approved):
I like!

PR #356, opened 2021-06-04
fix(tooltip): stop event propagation

gzimh:
References CAR-7544 (https://autoricardo.atlassian.net/browse/CAR-7544)

## Motivation and context
1. in some cases we forget and then do retroactive fixes. I'd like to find a solution that works out of the box without having to apply weird event handlers in every single tooltip
2. I'd like to understand better where the need for this is coming from. preferrably, we would always let the events bubble and not prevent them. there might be a fix in the tooltip library for this by now

## Before
Explicitly stop event propagation every place `Tooltip` is used.

## After
Out of the box solution. event propagation is stopped in the `Tooltip` component 

## How to test
* open a consumer app where the components pkg is used  (i.e Dealerhub)
* open listings page on Dealerhub
* make sure you're on Mobile
* open verified badge
* click the link
* it should simply open the link in a new tab (without navigating to the listing details in the original tab)

## Note
I don't think we have other options but to stop events bubble. Because ulimately it will bubble up into DOM hierarchy and reach the list item (in this specific case) which would redirect to listings detail page.

### DIFF
--- a/src/components/tooltip/index.tsx
+++ b/src/components/tooltip/index.tsx
@@ -38,7 +38,16 @@ const Tooltip: FC<Props> = ({
   alignment = TooltipAlignment.middle,
 }) => {
   return (
-    <div className="inline-block" onClick={(e) => e.preventDefault()}>
+    <div
+      className="inline-block"
+      onClick={(e) => {
+        e.preventDefault()
+        e.stopPropagation()
+      }}
+      onTouchEnd={(e) => {
+        e.stopPropagation()
+      }}
+    >
       <TooltipTrigger
         content={
           <div className="rounded-4 shadow-hard p-15 bg-white">
--- a/src/components/tooltip/index.tsx
+++ b/src/components/tooltip/index.tsx
@@ -38,19 +38,15 @@ const Tooltip: FC<Props> = ({
   alignment = TooltipAlignment.middle,
 }) => {
   return (
-    <div
-      className="inline-block"
-      onClick={(e) => {
-        e.preventDefault()
-        e.stopPropagation()
-      }}
-      onTouchEnd={(e) => {
-        e.stopPropagation()
-      }}
-    >
+    <div className="inline-block" onClick={(e) => e.preventDefault()}>
       <TooltipTrigger
         content={
-          <div className="rounded-4 shadow-hard p-15 bg-white">
+          <div
+            className="rounded-4 shadow-hard p-15 bg-white"
+            onTouchEnd={(e) => {
+              e.stopPropagation()
+            }}
+          >
             {renderContent()}
           </div>
         }


src/components/tooltip/index.tsx
```diff
@@ -41,7 +41,12 @@ const Tooltip: FC<Props> = ({
     <div className="inline-block" onClick={(e) => e.preventDefault()}>
       <TooltipTrigger
         content={
-          <div className="rounded-4 shadow-hard p-15 bg-white">
+          <div
+            className="rounded-4 shadow-hard p-15 bg-white"
+            onTouchEnd={(e) => {
```
talamcol, line 46: Do we understand why we need to do that onTouchEnd?

It concerns me a bit because the support is very bad on Safari. I quickly checked it on an iPhone 8 and the fix does not work. It is only officially supported after Safari 14.6 which is the newest version

https://caniuse.com/mdn-api_element_touchend_event

gzimh: @talamcol interesting. Well, this was a 'quick fix' that we added everytime we consumed the Tooltip component. So, the purpose of this PR was to centrealize it so we don't need to remember adding this everytime.

talamcol: @gzimh I think another way to solve this issue is to apply `pointer-event: all` to the wrapping `div`. The problem here is that the tooltip is absolute positioned and this should solve the issue.

Browser support looks better for this. But it's up to you, maybe let's hear another opinion

talamcol: Sorry for the confusion, it does only work the other way around. You can ignore my comment :)

masone: I agree with the sentiment though. let's try to understand what the initial fix does and why it was added exactly. is it about the tooltip being rendered on a card in a carousel? is there something about those touch/swipe events bubbling? I'd like to understand the underlying issue and solution better before we introduce it globally

one line above we also have a `e.preventDefault` on the click event. this is unusual as there's rarely a reason to prevent bubbling. can you dig a bit deeper before we settle for this workaround?

masone: would you like any support here?

gzimh: @masone i have some time today to dig a bit deeper into this. But i dont see a way other than stoping event propagation because the click/touch event will reach the Listing item and that will trigger it to redirect (just as an example, same thing would happen elsewhere). Maybe im missing something, will look into this a bit more

masone: I remember only the ontouch worked, but the click didn't that sounds weird!

masone: let's spend an hour and then proceed as-is. at least, it's tucked away nicely now

gzimh: On desktop it was working as it should (click). The onTouchEnd prevention is to fix the issue on mobile only

masone (approved):
as discussed in a call, we decided to move forward as-is. we can't explain why preventing the click is not enough. we tried.... it is what it is

PR #359, opened 2021-06-11
fix: badges behaviour for various usecases

lkappeler:
- fix: Only make the badge having a pointer cursor if there is a tooltip [ref](https://github.com/carforyou/carforyou-listings-web/pull/3246#issuecomment-858561423)
- fix: the colors for the verified and buyer protection badge for use cases like checkboxes in the filters [ref](https://github.com/carforyou/carforyou-listings-web/pull/3246#pullrequestreview-680884580)
- fix: make the bade background white in all uses cases [ref](https://github.com/carforyou/carforyou-listings-web/pull/3246#pullrequestreview-680677726)

### DIFF
--- a/src/components/badges/base.tsx
+++ b/src/components/badges/base.tsx
@@ -19,7 +19,7 @@ const BaseBadge: FC<Props> = ({ icon, text, size, tooltipContent }) => {
     <div
       className={classNames(
         styles.space,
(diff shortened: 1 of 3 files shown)


GutierrezGo (approved):
It looks good to me.

Thank you for the fixes 🌈

PR #367, opened 2021-06-22
New debounce

gzimh:


src/hooks/useDebounce.tsx
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from "react"
+
+const useDebounce = (onChange, debounce, initialValue) => {
+  const [event, setEvent] = useState(null)
+  const value = event?.target?.value || initialValue
```
masone, line 5: this looks dangerous. what happens if the user tries to set an empty value? I believe that would fall back to `initialValue`, no? please write a test for this case

src/hooks/useDebounce.tsx
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from "react"
+
+const useDebounce = (onChange, debounce, initialValue) => {
+  const [event, setEvent] = useState(null)
```
masone, line 4: why is the event in a state?

src/hooks/useDebounce.tsx
```diff
@@ -0,0 +1,16 @@
...
+  const value = event?.target?.value || initialValue
+  useEffect(() => {
+    if (event) {
```
masone, line 7: in which case would there be no event?

PR #368, opened 2021-06-23
feat: column segmented control

mreljin:
https://autoricardo.atlassian.net/browse/CAR-7787

Adds a segmented control that wraps to a column on `sm` resolutions. When there are >2 options with long texts, the basic segmented control tends to overflow horizontally on smaller screens.

related to: https://github.com/carforyou/carforyou-dealerhub-web/pull/1369

### DIFF
--- a/src/components/segmentedControl/button.tsx
+++ b/src/components/segmentedControl/button.tsx
@@ -10,6 +10,7 @@ interface Props {
   small?: boolean
   growing?: boolean
   position: "left" | "right" | "middle"
+  wrapOnSmallScreens?: boolean
   onClick?: () => void
 }
 
@@ -21,6 +22,7 @@ export const Button: FC<Props> = ({
   onClick,
   growing,
   position,
+  wrapOnSmallScreens,
 }) => {
   const padding = classnames("px-8", small ? "py-8" : "py-16")
   const { clonedElement, isWrapped } = wrapLink(children, padding)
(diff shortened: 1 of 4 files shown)


src/components/segmentedControl/button.tsx
```diff
@@ -21,28 +22,41 @@ export const Button: FC<Props> = ({
   onClick,
   growing,
   position,
+  wrapOnSmallScreens,
```
masone, line 25: I would prefer for this to be the default behavior on small screens. I prefer not to introduce code that accomodates for long text specifically on a case-by-case basis. we should design our components to work well out of the box, not by working around with specific props case-by-case :)

was this an explicit request from the designer? maybe we can talk to them and always have the stacking behavior on small viewports? this would relieve us from one more prop, all would display sanely by default and it's one less thing to test

masone: how would this behavior contrast to the growing control? 

1. what happens if I set both growing and wrapOnSmallScreens to true?
2. wouldn't the growing prop solve the issue with the width as well?

masone (changes requested):

src/components/segmentedControl/button.tsx
mreljin: > was this an explicit request from the designer?

No, I didn't want to change that across the entire app, so I went for changing it for this specific case, mostly because I didn't want to break the design in other places, if that makes sense. I agree with you that it would be easier if that was the default behavior. Who can I talk to for that?

masone: fair. I think it's something we should talk about with the designer and try to keep it as streamlined as possible with as little exceptions as possible :)

in this case, I would talk to Enrique. I think to remember he might be off soon, so I hope you can catch him today. let me know if you'd like me to join

mreljin:
Talked with Enrique and we agreed that the segmented control should not wrap and that it's better to use a radio group. Closing this.

PR #370, opened 2021-06-28
feat(debounce): lodash debounce

gzimh:
Ref: https://autoricardo.atlassian.net/browse/CAR-6929

This PR fixes the input debounce problems.

To use the debounce functionality for an input, just pass a prop debounce with a number as value (delay).

## Notes
In order to achieve this functionality as part of the `Input` component, it was neccessary to convert from controlled to uncontrolled inputs. 
Ref is used to access the input in order for the clear button to work


src/components/input/index.tsx
```diff
@@ -81,12 +85,13 @@ const Input = forwardRef<HTMLInputElement, Props>(
       min,
       max,
       position,
-      debounce,
+      debounce: delay,
```
masone, line 88: I know I commented on this before, but seeing how we're just aliasing it, it doesn't make sense in the full context. I'd piggy back from my previous request and name it `debounce` again. the code reads better with it actually

src/components/input/index.tsx
```diff
@@ -99,13 +104,39 @@ const Input = forwardRef<HTMLInputElement, Props>(
       visible: !!value,
       disabled,
       onClear: () => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ;(inputRef.current as any).value = ""
```
masone, line 108: is there a more concrete type than any that we can use? some lines down, we used `HTMLInputElement`?

src/components/input/index.tsx
```diff
@@ -99,13 +104,39 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
         onChange({ target: { name, value: "", cleared: true } })
       },
     }
 
+    const debouncedChangeHandler = useMemo(() => {
```
masone, line 113: what is the `useMemo` doing in this case?

src/components/input/index.tsx
```diff
@@ -99,13 +104,39 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
+      if (onChange) {
+        return debounce(onChange, delay)
+      }
+      return null
+    }, [])
+
+    useEffect(() => {
+      return () => {
+        debouncedChangeHandler?.cancel()
```
masone, line 122: that is a sweet upside from using the lodash function 👍

src/components/input/index.tsx
```diff
@@ -99,13 +104,39 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
+      }
+    }, [])
+
+    useEffect(() => {
+      if (ref) {
+        if (typeof ref === "function") {
+          ref(inputRef.current)
+        } else {
```
masone, line 130: can you shed light on the first branch of the conditional? when would ref not be a function? can we guard those cases with types instead of conditionals in the code?

masone (commented):
nice, that's big progress! thanks for digging through. this was really useful for understanding our components better. implementation looks solid to me - I'd like to understand the details a bit better, but we're close to merging I believe

src/components/input/index.tsx
gzimh: Well it was between `useMemo` and `useCallback` because we need to make sure the `debounce` instance is the same between re-renders. 
`useMemo` also calls debounce() only during initial rendering of the component, so yea, it provides a bit better performance.

src/components/input/index.tsx
gzimh: because refs can be passed as functions and objects.
```
myRef = React.createRef();
<>
  <input ref={this.myRef} />
  // Or
  <input ref={(ref) => (this.myRef.current = ref)} />
</>;
```
So, depending on what it's type is, we use a different way to assign its value

gzimh: https://github.com/Microsoft/TypeScript/issues/2072
I'ts not possible to use `instanceof` for `Function` type, it looks like its by design. 
```
typeof ref === "function"
```
doesn't look to bad  IMO though

src/components/input/index.tsx
gzimh: I think the alias is fine here because we import the loadash.debounce as `debounce` as well and this would avoid any confusions between the function and the delay time

masone: then we could go for naming the `prop` delay as well. that would be a breaking change, but given that it never worked.. I would not care much about that

src/components/input/index.tsx
```diff
@@ -99,13 +104,38 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
         onChange({ target: { name, value: "", cleared: true } })
       },
     }
 
+    const debouncedChangeHandler = useMemo(() => {
+      if (onChange) {
+        return debounce(onChange, delay)
+      }
+      return null
+    }, [])
```
masone, line 117: would we want to include onChange or delay in the deps? they'd never change in practice I guess...

src/components/input/index.tsx
masone: isn't the `ref` completely internal to this component? we can pretty much guarantee it's always one of the types, no?

src/components/input/index.tsx
```diff
@@ -124,15 +154,22 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
+        }
         onBlur={onBlur}
         onFocus={onFocus}
         onKeyDown={onKeyDown}
         autoFocus={autoFocus}
         step={step}
         min={min}
         max={max}
-        debounce={debounce}
+        debounce={delay}
```
masone, line 172: naming consistency again

src/components/input/index.tsx
gzimh: We didn't remove the forwardRef from the component, meaning that the consumer apps can still pass in `refs`. In those cases we combine the passed one with the internal one

src/components/input/index.tsx
gzimh: Yea, in practice those wont change, but its still nice to have them 👍

src/components/input/index.tsx
```diff
@@ -3,7 +3,12 @@ import React, {
   FocusEvent,
   forwardRef,
   KeyboardEvent,
+  MutableRefObject,
+  useEffect,
+  useMemo,
+  useRef,
 } from "react"
+import lodashDebounce from "lodash.debounce"
```
masone, line 11: naming problem solved 👍

src/components/input/index.tsx
masone: did you find any usage of the forwarded ref in the projects? last time I did a quick check, we didn't have one. so I'd consider dropping it so we have to support less branches

gzimh: I didn't find either. Sure, we can drop it

masone: we discussed to keep it

masone (approved):
please get another review and have QA test the forms with a pre-release

lkappeler (approved):
Looks good to me, the part with the reference is a bit counter-intuitive but it looks like you have discussed with Christian and it makes sense to me to support both versions

PR #372, opened 2021-06-28
fix: update form value if input filed is cleared

lkappeler:
Ref https://autoricardo.atlassian.net/browse/CAR-7632

If the selected date is cleared with the X in the input the callback was not executed, therefore the consuming component was not updated propperly

### DIFF
--- a/src/components/datePicker/index.tsx
+++ b/src/components/datePicker/index.tsx
@@ -91,10 +91,11 @@ const DatePicker: FC<Props> = ({
           scrollIntoViewIfMobile(containerRef.current, 300, false)
           setShowCalendar(true)
         }}
-        onChange={({ target: { value } }) => {
-          if (value === "") {
+        onChange={(event) => {
+          if (event.target.value === "") {
             setSelectedDate(null)
             setShowCalendar(false)
+            updateFormValue(event)
           }
         }}
         hasClearButton={!isToday(selectedDate)}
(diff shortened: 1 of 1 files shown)


GutierrezGo (approved):
I already checked the PR https://github.com/carforyou/carforyou-listings-web/pull/3332 in listings and it looks fine.

Thank you for the fix 🌈

Marine-Berthier (approved):

src/components/datePicker/index.tsx
```diff
@@ -73,6 +73,15 @@ const DatePicker: FC<Props> = ({
...
   }
 
+  const updateFormValue = (newValue) => {
+    callback({
+      target: {
+        name: formFiledName,
+        value: newValue,
+      },
+    })
+  }
```
masone, line 83: @gzimh inspiration related to your task with the controlled/uncontrolled delayed inputs. maybe this approach works for clearing in your case too and you don't have to fiddle with refs?

gzimh: While on a formik context the onChange its provided by it, the formik value gets updated to empty when the clear button is pressed but the input value is still left there because there is no way to update the value of an uncontrolled input (except via refs). 

I made this implementation with refs which does combine the incoming ref if it exists with an internal input one. So, even if the ref is passed from the consumer apps, they will still work fine (no breaking changes):
https://github.com/carforyou/carforyou-components-pkg/pull/370/commits/42f1c65639d0afcba2cd472cbd99c9cdfac2e664

PR #373, opened 2021-06-29
feat(collapsible): stop propagation when the target is anchor tag

gzimh:
References: https://autoricardo.atlassian.net/browse/CAR-7784
Stops event propagation in case the event's target is an anchor tag or it's child element of an anchor tag


src/lib/tagHelper.ts
```diff
@@ -0,0 +1,20 @@
...
+  tagName,
+  target,
+  currentTarget,
+}: {
+  tagName: string
+  target: HTMLElement
+  currentTarget: HTMLElement
+}) => {
+  if (target == currentTarget || !target.parentElement) return false
+  if (target.tagName.toLowerCase() === tagName.toLowerCase()) return true
```
masone, line 11: any chance target.tagName could be undefined?

src/components/collapsible.tsx
```diff
@@ -52,9 +55,17 @@ const Collapsible: FC<Props> = ({
...
         })}
-        onClick={() => {
-          onChange && onChange(!isCollapsed)
-          setIsCollapsed(!isCollapsed)
+        onClick={(e) => {
+          if (
+            !hasEventBubbledThroughTag({
+              tagName: "a",
+              target: e.target as HTMLElement,
+              currentTarget: e.currentTarget,
```
masone, line 63: any chance target/currentTarget could be undefined?

src/lib/tagHelper.ts
```diff
@@ -0,0 +1,20 @@
+const hasEventBubbledThroughTag = ({
+  tagName,
+  target,
+  currentTarget,
```
masone, line 4: consider passing the event over target/currentTarget separately

src/lib/tagHelper.ts
```diff
@@ -0,0 +1,20 @@
+const hasEventBubbledThroughTag = ({
```
masone, line 1: if you only use this function in one place, I'd not introduce a helper for it just yet, but bundle the function with the component

src/components/collapsible.tsx
gzimh: I don't think so, each onClick event should have a target and a currentTarget

src/lib/tagHelper.ts
gzimh: The reason why I send those as separate params is beacuse this is a recursive function and in the subsequent calls, for the target prop we send the target's parent.

src/lib/tagHelper.ts
gzimh: Even though I don't think that's possible because that would be an invalid html tag, I added this check to make sure: https://github.com/carforyou/carforyou-components-pkg/pull/373/files#diff-6023195695dc84b77588cf36adfcd004d5cc40ce91e778fbb233e0f84bb94ed7R60

masone (approved):
looks good to me. maybe let's get an approval from @Averethel as well - he once did similar things in the `Link`-ish component if my memory does not fail me

Averethel (approved):
Looks good to me. 
Would be great if we could test that functionality too.

PR #381, opened 2021-07-07
feat: add image component

Marine-Berthier:
References [CAR-7455](https://autoricardo.atlassian.net/browse/CAR-7455?atlOrigin=eyJpIjoiYjU5NGE4ZTE5MDY0NDA5ZjlhODM3M2UwNjk4ZmMzZjIiLCJwIjoiaiJ9)

[PR for listings](https://github.com/carforyou/carforyou-listings-web/pull/3364)
[PR for DH](https://github.com/carforyou/carforyou-dealerhub-web/pull/1400) = to be updated


src/components/image.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { FC } from "react"
+
+interface Props {
+  image: string
+  className?: string
+  onLoad?: () => void
+  getImageAltAttribute?: string
```
lkappeler, line 7: In terms of naming, it sounds like a function that is passed but it's just the attribute. Can we simply name it `alt` in this case?

src/components/image.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { FC } from "react"
+
+interface Props {
+  image: string
+  className?: string
```
lkappeler, line 5: A while ago we discovered that it is kind of an anti-pattern to pass class names to a component as it allows for a lot of variation. Respectively under certain circumstances, the appearance of the component can be completely different than what we intended in the components package. Instead of allowing to pass classes, we decided to identify the variations that the component should provide and define an interface for that. See for [example](https://carforyou.dev/carforyou-components-pkg/?path=/story/button--default) the `size` and `style` attribute on the button, that allows for a set of combinations but prevents the consumer from doing something like a yellow button with purple text. For the Image I would expect something like `object-fit` to be a potential variation. If you want to we can take a look together on Monday and define how the interface should look that we wanna provide for our images.

lkappeler (changes requested):

src/components/image.tsx
Marine-Berthier: Thank you for the feedback @lkappeler. I worked on an interface today, it is pretty straighforward to make it work for DH but more complicated for listings. I will try to fix my issues in listing tomorrow morning with a clear mind but I might ask for help indeed.

src/components/image.tsx
```diff
@@ -0,0 +1,58 @@
...
+  objectFit?: "cover" | "contain"
+}
+
+const Image: FC<Props> = ({ image, objectFit, imageType, onLoad, alt }) => {
+  const cover = classnames({
+    "object-cover": objectFit === "cover",
+    "object-contain": objectFit === "contain",
+  })
+
+  const classes = classnames({
```
Marine-Berthier, line 28: @lkappeler When you suggested to create an interface you were thinking about something like this right ? 
It looks a bit messy since there's a lot of variations. Would you advice to split it again like I did for objectFit : 1 props for `width`, another one for `height` and then the `classes` ?

src/components/image.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { FC } from "react"
+import classnames from "classnames"
+
+interface Props {
+  image: string
```
Averethel, line 5: Naming-wise and the way it's used I'd call it `src` or `source` if you want to be more verbose

src/components/image.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <img
+      data-testid="component-image"
+      src={image}
+      alt={alt}
+      onLoad={onLoad}
+      onError={(e) => {
+        onError && onError(e)
+      }}
```
Averethel, line 38: Wouldn't this work:
```suggestion
      onError={onError}
```
?

Or is something going wrong if you pass an `undefined` as an event handler?

src/components/image.tsx
```diff
@@ -0,0 +1,51 @@
...
+      className={
+        galleryImage
+          ? `${cover} m-auto max-h-screen h-full`
+          : `${cover} h-full w-12/12`
+      }
```
Averethel, line 43: You're passing cover in both branches. I'd go for:

```suggestion
      className={classnames(cover, 
        galleryImage
          ? "m-auto max-h-screen h-full"
          : "h-full w-12/12"
      )}
```

src/stories/image.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+const Template: FC<Props> = (args) => {
+  return (
+    <StoryContainer
+      title={args.storyName}
+      component={<Image {...args}>{args.children}</Image>}
+    />
+  )
+}
+
+export const Default = Template.bind({})
```
Averethel, line 26: There are quite some more variations of the image. Can we showcase more here?

src/components/image.tsx
Marine-Berthier: No it works fine as you suggested !

src/stories/image.stories.tsx
Marine-Berthier: We can but there will be no visible difference in Storybook

Averethel (approved):

src/tailwind/defaultConfig.ts
```diff
@@ -362,9 +364,13 @@ export default {
     |*/
 
     maxHeight: {
+      screen: "100vh",
       full: "100%",
       dropdown: "240px",
+      image: "180px",
+      imageLarge: "330px",
       dropdownSM: "190px",
+      imageInfo: "400px",
```
masone, line 373: 🤔 can we keep these sizes in the projects? they're not referenced within the image component at all. it feels like a breach of separation of concerns to me. the component library should not know anything about the sizes used in projects and should be flexibly working with any

even if it's just about sharing common sizes, I'd keep them in the projects in this case as it's a layout concern, not a component concern

src/components/image.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <img
+      data-testid="component-image"
+      src={source}
+      alt={alt}
+      onLoad={onLoad}
+      onError={onError}
+      className={classnames(
+        cover,
+        galleryImage ? "m-auto max-h-screen h-full" : "h-full w-12/12"
```
masone, line 39: `h-full` is shared between both variants :)

src/components/image.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const Image: FC<Props> = ({
+  source,
+  objectFit,
+  onLoad,
+  alt,
+  width,
+  height,
+  onError,
+  galleryImage,
```
masone, line 23: can we find a name that imposes less how the image is used and more on what the effect of the prop will be? is it `fullSize` or `fullScreen` or something?

src/stories/image.stories.tsx
masone: `galleryImage` and `cover` should make a difference. you might need some additional markup around the image in the story (which is fair game) to demo the latter

Marine-Berthier: You're right ! With additional style to the StoryContainer we can clearly show `cover`, could not make a huge difference with` galleryImage` though (which is called `fullScreenHeight` now). I will try again tomorrow.

src/tailwind/defaultConfig.ts
Marine-Berthier: I get what you mean, I removed them and I will adjust the tailwind config directly in listings and DH

src/components/image.tsx
Marine-Berthier: I went for `fullScreenHeight` but `fullScreen` might be enough actually 👍

lkappeler (approved):
Nice work, looks very clean with the interface now from what we started off with 👍
thanks for pushing through with this, huge effort 🎉

PR #382, opened 2021-07-08
chore: clean up now that npm link works

masone:
https://github.com/carforyou/carforyou-listings-web/pull/3360


Averethel:
@masone I think there is a misconception. We still need a custom link script to ensure that the version of react is linked back to the package, but that was broken since the linked package wasn't piped through the babel loader.

gzimh:
> @masone I think there is a misconception. We still need a custom link script to ensure that the version of react is linked back to the package, but that was broken since the linked package wasn't piped through the babel loader.

True, we still need the bash script

masone:
oh my. gotcha. so it's more workarounds rather than less ;)

gzimh:
Well, if we don't use this bash script we are still fine, but we need to run `npm link` in the components pkg first, then in the consumer project we need to run `npm link <path_to_pkg>`. So the script is doing this for us instead.

PR #384, opened 2021-07-12
Extend Pagination component

gzimh:
https://autoricardo.atlassian.net/browse/CAR-6910 (.15 - .17)

This adds a way for the customer apps to be notified on exatly what pagination button has been clicked: 
* previous
* next
* page (number) button (i.e `2`)

Note: The pagination library used `react-paginate` doesn't add support for this functionality out of the box, so, we did a custom logic that takes care of it.

### DIFF
--- a/src/components/pagination.tsx
+++ b/src/components/pagination.tsx
@@ -1,5 +1,5 @@
 import ReactPaginate from "react-paginate"
-import React, { StatelessComponent } from "react"
+import React, { StatelessComponent, useEffect, useState } from "react"
 
 import ArrowDownM from "./icons/arrowDownMCrop"
 
@@ -13,6 +13,7 @@ interface Props {
   query?: Record<string, unknown>
   pageLinkBuilder?: (page: number, params: Record<string, unknown>) => string
   renderHead?: (links) => void
+  onButtonClick?: (label) => void
 }
 
 const getRelLinks = (forcePage, query, pageCount, linkBuilder) => {
@@ -47,7 +48,32 @@ export const Pagination: StatelessComponent<Props> = ({
   query,
   pageLinkBuilder,
   renderHead,
+  onButtonClick,
 }) => {
+  const [currentPage, setCurrentPage] = useState(null)
+  const [navButtonClicked, setNavButtonClicked] = useState("")
+
+  const handlePageChange = (data) => {
+    setCurrentPage(data)
+  }
+
+  const handleNavButtonClick = (label) => {
+    setNavButtonClicked(label)
+  }
+
+  useEffect(() => {
+    if (currentPage) {
+      if (navButtonClicked && onButtonClick) {
+        onButtonClick(navButtonClicked)
+      } else if (currentPage && onButtonClick) {
+        onButtonClick(`${currentPage.selected + 1}`)
+      }
+      onPageChange(currentPage)
+      setNavButtonClicked("")
+    }
+  }, [currentPage])
+
+  useState()
   return pageCount > 1 ? (
     <div className="flex justify-center">
       {renderHead &&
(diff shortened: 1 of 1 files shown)


src/components/pagination.tsx
```diff
@@ -47,7 +48,32 @@ export const Pagination: StatelessComponent<Props> = ({
...
+
+  const handlePageChange = (data) => {
+    setCurrentPage(data)
+  }
+
+  const handleNavButtonClick = (label) => {
+    setNavButtonClicked(label)
+  }
+
+  useEffect(() => {
```
ufkaya, line 64: Do we really need this useEffect? Cant we just call the functions inside `handlePageChange` and `handleNavButtonClick`?

gzimh: Yea we actually needed. It's a bit of a complex case becasue we have a way to tell if `next/prev` buttons have been clicked, but not if pagination buttons (numbers) have been clicked.
Basically, when `next/prev` is called, we just update a state with that label. That same click will also trigger the value (page) state to change. That's why the only way I was able to make this work was via the useEffect hook.

ufkaya: can you use only the `handlePageChange`? So we can remove the currentPage state

ufkaya (approved):

src/components/pagination.tsx
gzimh: I know it seems intuitive but the issue with that is because the same click event of nav/prev and page number buttons also triggers the pagination's state to change. 
So, in order to create some 'delay' to give enough time for `navButtonClicked` to be populated with the label, we need the useffect to be triggered on state change and then pass that new state to the consumers

src/components/pagination.tsx
```diff
@@ -47,7 +48,32 @@ export const Pagination: StatelessComponent<Props> = ({
...
+    if (currentPage) {
+      if (navButtonClicked && onButtonClick) {
+        onButtonClick(navButtonClicked)
+      } else if (currentPage && onButtonClick) {
+        onButtonClick(`${currentPage.selected + 1}`)
+      }
+      onPageChange(currentPage)
+      setNavButtonClicked("")
+    }
+  }, [currentPage])
```
talamcol, line 74: you are missing some dependencies of the hook here

src/components/pagination.tsx
ufkaya: Which function is first called? If it is the `handleNavButtonClick ` then you can also use the `useRef()` to store the data

src/components/pagination.tsx
gzimh: Already discussed with @talamcol in slack about this one. We decided that its best to keep the deps array as is in this use case.

masone:
what's the motivation for this change? can you link a related ticket?

src/components/pagination.tsx
```diff
@@ -13,6 +13,7 @@ interface Props {
   query?: Record<string, unknown>
   pageLinkBuilder?: (page: number, params: Record<string, unknown>) => string
   renderHead?: (links) => void
+  onButtonClick?: (label) => void
```
masone, line 16: from an interface-perspective, I would not care what underlying elements are used. shall we simplify to 
```suggestion
  onClick?: (label) => void
```

or is it worth being so specific? maybe there's something I don't see...

masone (commented):
write some tests for it please - there are nested conditionals with multiple branches in here

gzimh:
> what's the motivation for this change? can you link a related ticket?

https://autoricardo.atlassian.net/browse/CAR-6910
Nr .15 to .17

src/components/pagination.tsx
gzimh: Yea, `onClick` fits better indeed

masone:
Thanks. if this implementation proves to be too problematic, I'd be happy to skip these trackings. We don't have them for any other pagination, so I would not invest too much time in making it happen for a sideshow such as printcenter. If you can't fulfil the requirement in another 1-2h, let's step back from this. Let me know if you need backup discussing this with the PO

gzimh:
> Thanks. if this implementation proves to be too problematic, I'd be happy to skip these trackings. We don't have them for any other pagination, so I would not invest too much time in making it happen for a sideshow such as printcenter. If you can't fulfil the requirement in another 1-2h, let's step back from this. Let me know if you need backup discussing this with the PO

I agree. To much workarounds for a simple tracking. We can always track the pagination (page number) on change though. I think that is good enough and there is little to no value in detecting if the `next/prev` buttons specifically have been used to change the pagination. 

@masone

masone:
Agreed. let's ditch this and go with what we have. it's likely not a hard requirement. send POs my way if they approach you about it

gzimh:
We decided to close this PR. The workaround is not worth it in this case.

PR #388, opened 2021-07-13
fix: update tooltip library to allow update to react 17

talamcol:


package.json
```diff
@@ -36,7 +36,7 @@
     "glob": "^7.1.6",
     "react-calendar": "^3.3.1",
     "react-paginate": "^7.0.0",
-    "react-tooltip-lite": "^1.12.0",
+    "react-tooltip-lite": "npm:@carforyou/react-tooltip-lite@1.12.1",
```
masone, line 39:
```suggestion
    "@carforyou/react-tooltip-lite": "1.12.1",
```

masone (approved):

package.json
talamcol: this does not work because the `declare module` is named differently :D

masone: uuuh. then I fucked up with the release. this should be very temporary, so let's not sweat it

PR #390, opened 2021-07-15
fix: adjust title and close on modal

GutierrezGo:
> Jira https://autoricardo.atlassian.net/browse/CAR-8086

## Motivation and context
The issue could be fixed individually and faster in listings but we can face the same issue in the future. Therefore, the fix lies in the modal component.

## How to test
Please check the adjustments in DH and listings before you approve this PR:

**DH:** https://github.com/carforyou/carforyou-dealerhub-web/pull/1426
**Listings:** https://github.com/carforyou/carforyou-listings-web/pull/3388

# Thank you 🌯

### DIFF
--- a/src/components/modal/index.tsx
+++ b/src/components/modal/index.tsx
@@ -23,7 +23,7 @@ export interface ModalProps {
   // Vertical overflow
   verticalOverflow?: ModalOverflow
   // Title in the modal
-  title?: ReactNode
+  title?: string
 }
 
 const Modal: FC<ModalProps> = ({
@@ -64,7 +64,7 @@ const Modal: FC<ModalProps> = ({
       onClick={handleClose}
       data-testid="modal-close"
     >
-      <CloseMIcon color={style === "dark" ? "#FFFFFF" : "#232A36"} />
+      <CloseMIcon />
     </div>
   )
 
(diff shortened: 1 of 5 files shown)


src/components/modal/index.tsx
```diff
@@ -22,6 +22,8 @@ export interface ModalProps {
   style: ModalStyle
   // Vertical overflow
   verticalOverflow?: ModalOverflow
+  // Title in the modal
+  title?: ReactNode
```
masone, line 26: do we have different sizes/styles of titles or could we just go with a text prop? that would allow us to harmonize the title without duplicating any markup in the projects

src/components/modal/index.tsx
```diff
@@ -81,6 +77,28 @@ const Modal: FC<ModalProps> = ({
...
+              </div>
+            </div>
+          ) : (
+            <div
+              className="absolute z-modalClose cursor-pointer right-modalClose top-modalClose"
+              onClick={handleClose}
+              data-testid="modal-close"
+            >
+              <CloseMIcon color={style === "dark" ? "#FFFFFF" : "#232A36"} />
+            </div>
```
masone, line 100: the whole DOM for the close is duplicated, can you share it between the two branches of the condition? :)

src/components/modal/index.tsx
GutierrezGo: In all the modals in listings and DH the markup is:
`<h2 className="font-bold text-xl text-grey-dark w-12/12">` So, now we are not duplicating that markup. We pass only the translation.

GutierrezGo:
Christian comment -> https://github.com/carforyou/carforyou-listings-web/pull/3386#issuecomment-881465133

src/components/modal/index.tsx
GutierrezGo: I have changed the type of the prop to a `string`

lkappeler (approved):
LGMT, thanks for the cleanup 🙂

PR #391, opened 2021-07-16
CAR-8098: Integrate exhaustive-deps linting rule for pkgs in Components PKG

gzimh:
References [CAR-8098](https://autoricardo.atlassian.net/browse/CAR-8098)

## Motivation and context

We were working on the task to specify all missing useEffect dependencies. It is completed for listings and dealerhub. However, the rule is also important in our pkgs as well.

## Before

No linting rule and missing dependencies

## After

Linting rule and all dependencies are specified

## How to test

Browse through the apps where the `components-pkg` is used and make sure everything still works as expected. Keep an eye on:
- inputs
- modals
- intercom


src/components/intercom/index.tsx
```diff
@@ -84,23 +84,27 @@ export const Intercom: FC<Props> = ({
...
+      stage,
+      custom_launcher_selector: `#${intercomLauncherId}`,
+      hide_default_launcher: true,
+      alignment: "right",
+      horizontal_padding: 20,
+      vertical_padding: 76,
+      language_override: language,
+      ...userInfo,
+    }
+  }, [appId, language, stage, userInfo])
```
talamcol, line 100: user info is an object, `useMemo` only compares the reference which would result in your memoized object being recreated on every component change

src/components/intercom/index.tsx
```diff
@@ -123,7 +127,7 @@ export const Intercom: FC<Props> = ({
       window.Intercom("shutdown")
       bootIntercom(intercomSettings, intercomEventHandlers)
     }
-  }, [isLoggedIn])
+  }, [intercomEventHandlers, intercomSettings, isLoggedIn])
```
talamcol, line 130: `intercomSettings` is an object. Thus you need `useDeepCompare` here

src/components/modal/index.tsx
```diff
@@ -35,22 +35,25 @@ const Modal: FC<ModalProps> = ({
   const overflowClass = verticalOverflow
     ? `overflow-y-${verticalOverflow}`
     : "md:overflow-y-auto overflow-y-scroll"
-  const handleClose = () => {
+  const handleClose = useCallback(() => {
     onClose ? onClose() : null
     close()
-  }
+  }, [onClose, close])
```
talamcol, line 41: did you make sure that every component that uses the modal and passes a `onClose` function wraps it into a callback? because otherwise this function is recreated on every change of the parent. Thus triggers a change on handleClose, then handleKeys and at last the useEffect

src/components/input/index.tsx
```diff
@@ -120,7 +120,7 @@ const Input = forwardRef<HTMLInputElement, Props>(
       return () => {
         debouncedChangeHandler?.cancel()
       }
-    }, [])
+    }, [debouncedChangeHandler])
```
talamcol, line 123: debouncedChangeHandler changes everytime the `onChange` function changes. As a result, you should check if all components have this function in a callback

src/components/modal/index.tsx
gzimh: For simple components, I saw that it's fine to let the component rerender because in most cases its a lighter operation than keeping a cache / checking against it.  That's basically the reason we shoudn't wrap every function with `useCallback`

src/components/input/index.tsx
talamcol: have you checked if the number of times the hook runs is fine?

.eslintrc.js
```diff
@@ -8,5 +8,17 @@ module.exports = {
         "import/no-internal-modules": "off",
       },
     },
+    {
+      files: ["*"],
+      rules: {
+        "no-restricted-imports": "off",
```
talamcol, line 14: I think this rule is a copy/paste error

talamcol:
changes look good to me if the above comments are addressed. Maybe you can create a branch for listings/dealerhub and include the prerelease there to run it through QA. Since it involves the modal, intercom and input it might have a bigger impact

src/components/input/index.tsx
talamcol: looks okeyish

src/components/intercom/index.tsx
talamcol: looks okeyish

src/components/input/index.tsx
ufkaya, line 123: Alternatively, we can also use `useDeepCompareEffect`

src/components/intercom/index.tsx
```diff
@@ -84,31 +84,35 @@ export const Intercom: FC<Props> = ({
...
-    stage,
-    custom_launcher_selector: `#${intercomLauncherId}`,
-    hide_default_launcher: true,
-    alignment: "right",
-    horizontal_padding: 20,
-    vertical_padding: 76,
-    language_override: language,
-    ...userInfo,
-  }
+  const intercomSettings = useMemo(() => {
```
ufkaya, line 87: Do we really need `useMemo`? We can also remove it because we are using `useDeepCompareEffect`

src/components/modal/index.tsx
```diff
@@ -38,22 +38,25 @@ const Modal: FC<ModalProps> = ({
...
+      }
+    },
+    [handleClose]
+  )
 
   useEffect(() => {
     document.addEventListener("keydown", handleKeys)
     return () => document.removeEventListener("keydown", handleKeys)
-  }, [])
+  }, [handleKeys])
```
ufkaya, line 59: Alternatively, we can also use `useDeepCompareEffect`

src/components/modal/index.tsx
ufkaya: if the `onClose` will always have a different ref then we dont need `useCallback`

ufkaya (approved):
LGTM

src/components/intercom/index.tsx
talamcol: yes I think we need it because `intercomSettings` depends on props. Otherwise the object would be recreated on every render cycle

src/components/modal/index.tsx
talamcol: handleKeys is a function

talamcol: as discussed, we will disable the rule for this component

src/components/modal/index.tsx
talamcol: as discussed, we will disable the rule for this component

src/components/input/index.tsx
talamcol: as discussed, we will disable the rule for this component

PR #392, opened 2021-07-19
fix: pin version for pika-plugin-build-components

acapop77:
chore: pin pika-plugin-build-components to version 1.0.10 in order to safely test updating is dependency packages


masone:
remember that the chore will not release the package. if you need a new release, do a fix instead

PR #393, opened 2021-07-20
CAR-8045: new style "white" for button

talamcol:
Reference: [CAR-8045](https://autoricardo.atlassian.net/browse/CAR-8045)

For the new make page, we need a white button:
image

I refactored the styles a bit and went for more duplication. It was no longer readable when adding another style

### DIFF
--- a/src/components/button.tsx
+++ b/src/components/button.tsx
@@ -7,7 +7,7 @@ interface Props {
   /** - */
   children: ReactNode
   dataTestid?: string
-  style?: "salmon" | "teal" | "teal-border"
+  style?: "salmon" | "teal" | "teal-border" | "white"
   size?: "large" | "small" | "responsive"
   disabled?: boolean
   onClick?: (event: MouseEvent<HTMLButtonElement>) => void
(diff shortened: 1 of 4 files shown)


src/components/button.tsx
```diff
@@ -7,7 +7,7 @@ interface Props {
   /** - */
   children: ReactNode
   dataTestid?: string
-  style?: "salmon" | "teal" | "teal-border"
+  style?: "salmon" | "teal" | "teal-border" | "white"
```
GutierrezGo, line 10: What do you think about `white-border`? then we apply the same understanding that we have for the `teal-border`

GutierrezGo:
Is it possible to add a test for these variations?

src/components/button.tsx
```diff
@@ -37,17 +37,29 @@ export const Button: FC<Props> = ({
...
+    teal: "bg-teal hover:bg-teal-dark focus:bg-teal border-teal",
+    "teal-border": "bg-white text-teal hover:opacity-60 border-teal",
+    white: "bg-none text-white hover:opacity-60 border-white",
+    salmon: "bg-salmon border-salmon hover:bg-salmon-dark focus:bg-salmon",
+  }
+
+  const disabledStyles = {
+    teal: "bg-grey-1 border-grey-1 hover:bg-grey-1 text-grey-4",
+    "teal-border": "text-grey-3 border-grey-3 hover:opacity-100",
+    white: "text-grey-3 border-grey-3 hover:opacity-100",
```
GutierrezGo, line 51: Is it possible to make a clean up here instead of repeating the same styles?

talamcol:
> Is it possible to add a test for these variations?

there is already a snapshot test for two cases

src/components/button.tsx
talamcol: I duplicated on purpose because personally I think it's more visible what happens when we duplicate it. Before it was already quite complex with a lot of conditions

masone (approved):
don't forget the semantic commit message 🤞

PR #394, opened 2021-07-21
fix: padding in gallery modal

GutierrezGo:
Accidentally I have moved the paddings for all the sizes but the fullscreen shouldn't have those paddings. Otherwise, it will be padding in the gallery modal in listings on the PDP and as a result a horizontal scrollbar.

# Thank you 🔪


Marine-Berthier (approved):

masone (approved):

masone:
glad you noticed!

PR #395, opened 2021-07-22
fix: disable autocomplete

ufkaya:
Chrome ignores `off`, on my browser, chrome version  91, works with `none`


Averethel (changes requested):
Wouldn't this result in invalid markup? IIRC valid attributes are `on`, `off` and some hints on which field this field referring to (in case of the address, CC etc.).

Did you consider adding `autocomplete=off` on the form level? I recall at some point this was needed for it to work in chrome. 

Do we know why it doesn't work? Is it a bug in chrome or was it a decision of the chrome team to ignore the attribute?

If we go this route what are the implications for the other browsers? Would it still work? What about the older versions?

Sorry, it's a bit too many questions to not block the PR for now.

ufkaya:
> Did you consider adding `autocomplete=off` on the form level? I recall at some point this was needed for it to work in chrome.

good point tried and seems to work but the filters are not inside a form, I will close this PR and we can create a new task

PR #396, opened 2021-07-27
fix: handle null gbdScore gracefully

Averethel:
Due to changes in the calculation of `gdbScore`, there is no default anymore. That means `null` can be sent when the calculation is not yet finished on the BE side. We've been asked to handle `nulls` the same way as `not-defined` in the frontend since it's not feasible to set the default value on elasticsearch.

###  Before
```
<GdbBadge score={null} />
```
Would render a black, unstyled label without text

### After
```
<GdbBadge score={null} />
```
looks the same as
```
<GdbBadge score={"not-defined"} />
```

### DIFF
--- a/src/components/badges/__tests__/gdbd.test.tsx
+++ b/src/components/badges/__tests__/gdbd.test.tsx
@@ -34,7 +34,7 @@ describe("<GdbdBadge>", () => {
       expect(getByText("Above market price"))
     })
 
-    it("renders the badge fon not defined", () => {
+    it("renders the badge for not defined", () => {
       const { getByText } = render(
         <GdbdBadge language="en" score="not-defined" />
       )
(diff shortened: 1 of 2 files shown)


src/components/badges/__tests__/gdbd.test.tsx
```diff
@@ -34,12 +34,17 @@ describe("<GdbdBadge>", () => {
       expect(getByText("Above market price"))
     })
 
-    it("renders the badge without a score", () => {
+    it("renders the badge fon not defined", () => {
```
ufkaya, line 37:
```suggestion
    it("renders the badge for not defined", () => {
```

ufkaya (approved):

talamcol (approved):

PR #406, opened 2021-08-13
fix: migrate to assets-pkg

masone:
https://autoricardo.atlassian.net/browse/CAR-8227

generating assets based on our conventions


package.json
```diff
@@ -76,15 +74,12 @@
     "postcss-easy-import": "^3.0.0",
     "postcss-loader": "^4.2.0",
     "react": "^16.14.0",
-    "react-docgen-typescript-loader": "^3.6.0",
```
masone, line 79: part of the latest storybook version

package.json
```diff
@@ -76,15 +74,12 @@
...
-    "react-dom": "^16.12.0",
-    "react-test-renderer": "^16.12.0",
```
masone, line 81: was not really needed, built into testing-lib

package.json
```diff
@@ -76,15 +75,12 @@
     "postcss-easy-import": "^3.0.0",
     "postcss-loader": "^4.2.0",
     "react": "^16.14.0",
```
GutierrezGo, line 77: just one open question. Is it possible to upgrade the react version as in the other projects?

GutierrezGo (commented):
Sorry, one question do we need to move the icons into the assets/src as well?

masone:
Thanks, I missed those! Will move them as well and ping you again for review

GutierrezGo (approved):
It looks good to me.
Thanks!! 🌈

talamcol (approved):
Looks good!

PR #409, opened 2021-08-24
feat: do not rely on input being required to display the floating label

lkappeler:
Ref: https://autoricardo.atlassian.net/browse/CAR-8253

For the migrated buyer protection form we wanna use the HTML validation in the first version.
The floating label input makes every input required which prevents us from doing so.
Instead of using the `:valid` pseudo selector use the `:placeholder-shown` to make the label floating

Testing:

Check out the branch and ensure the floating labels are working the same as before

### DIFF
--- a/assets/input/floatingLabel.css
+++ b/assets/input/floatingLabel.css
@@ -8,17 +8,18 @@ input.floatingLabel-input:focus {
 }
 
 input.floatingLabel-input:focus ~ .floatingLabel-label {
-  @apply text-teal
+  @apply text-teal;
 }
 
 .floatingLabel-input[data-valid="false"] ~ .floatingLabel-label,
 .floatingLabel-input[data-valid="false"]:focus ~ .floatingLabel-label {
-  @apply text-salmon
+  @apply text-salmon;
 }
 
 .floatingLabel-input:focus ~ .floatingLabel-label,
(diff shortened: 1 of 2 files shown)


assets/input/floatingLabel.css
```diff
@@ -8,31 +8,44 @@ input.floatingLabel-input:focus {
 }
 
 input.floatingLabel-input:focus ~ .floatingLabel-label {
-  @apply text-teal
+  @apply text-teal;
```
djanov, line 11: not sure why is ` ;` added

assets/input/floatingLabel.css
```diff
@@ -8,31 +8,44 @@ input.floatingLabel-input:focus {
...
 }
 
 .floatingLabel-input[data-valid="false"] ~ .floatingLabel-label,
 .floatingLabel-input[data-valid="false"]:focus ~ .floatingLabel-label {
-  @apply text-salmon
+  @apply text-salmon;
```
djanov, line 16: same here

djanov (approved):
LGTM, thank you 💯

assets/input/floatingLabel.css
lkappeler: My Ide decided to as it was there on all other lines I thought I would keep it for consistency

PR #410, opened 2021-09-03
feat: allow to hide the close icon on modals

talamcol:


masone:
that's not a feature, that's a dark pattern! :D
why do we make it harder to close modals? what's our use case for this? :)

talamcol:
Will solve it different :)

PR #412, opened 2021-09-06
feat: add prop to pass a custom testId for collapsibles

lkappeler:
Allow passing a custom test-id
During the hackday project of dynamic filters, I extracted the collapsible filters to a component in order to add a guard which leads to issues with the integration test. Passing a custom test-id should allow to directly select the required collapsible and avoid looking for text that may occur multiple times in the filters

ref: https://github.com/carforyou/carforyou-listings-web/pull/3537


GutierrezGo (approved):
🌈

src/components/collapsible.tsx
```diff
@@ -32,6 +32,8 @@ interface Props {
   forceCollapse?: boolean
   /** Makes sure that the content is always in the DOM */
   alwaysRender?: boolean
+  /** Overrides the generic test-id "collapsible" with the passed prop*/
+  dataTestId?: string
```
masone, line 36: Personal opinion ahead :) 

I dislike to add props for testing purposes and tend to be a bit more picky in components-pkg than in project code. Is there really no other way? Can we instead query in the test for text in the collapsible? Can we add a custom testid to the parent and query for that instead?

masone: this thing already has like 5 optional props, which starts to smell too...

lkappeler: So, my thought process behind was there is a hardcoded property inside the component for testing purposes so rather make it configurable if the consumer needs to.

Don't have a strong opinion, likely I can workaround adding it and make the tests work. It will make the tests more complex and nested. The thing about the text is that the category selector is not unique anymore...

PR #413, opened 2021-09-06
fix: remove svgs from api

masone:
BREAKING CHANGE: stops exposing svgs from components-pkg

the assets in question have been encapsulatd in header-footer. they're used in no project


GutierrezGo (approved):
It looks good to me thanks 😊

PR #419, opened 2021-09-13
feat: allow images to lazy load

masone:
https://autoricardo.atlassian.net/browse/CAR-8495

### DIFF
--- a/src/components/image.tsx
+++ b/src/components/image.tsx
@@ -10,6 +10,7 @@ interface Props {
   width?: number
   height?: number
   onError?: (e) => void
+  lazy?: boolean
 }
 
 const Image: FC<Props> = ({
@@ -21,6 +22,7 @@ const Image: FC<Props> = ({
   height,
   onError,
   fullScreen,
+  lazy = false,
 }) => {
   const cover = classnames({
     "object-cover": objectFit === "cover",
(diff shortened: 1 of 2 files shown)


src/components/image.tsx
```diff
@@ -10,6 +10,7 @@ interface Props {
   width?: number
   height?: number
   onError?: (e) => void
+  lazy?: boolean
```
lkappeler, line 13: In terms of interface, why adding a facade instead of using loading lazy and eager directly? To me, that would be more straightforward to understand as it's a common standard and I don't have to look up how CFY is mapping lazy...

masone: my thinking was: eager is the default, so it ends up being a boolean toggle

masone: (I'm open to changing it)

lkappeler: We still proxy the attribute the property we are setting is `loading`. I'll leave it up to you, don't have that strong of an opinion.

lkappeler (approved):

Marine-Berthier (approved):

src/components/image.tsx
talamcol: I'm also with Lorenzo on the interface as boolean props can facilitate confusion and wrong usage

talamcol (approved):
looks good to me but I would also change the interface to a string type (same as the normal img)

masone:
Thanks for sharing your opinions. Now that we have a third opinion, I'll change the interface :)

PR #420, opened 2021-09-15
fix: info icon size for label popup

masone:
https://autoricardo.atlassian.net/browse/CAR-8536

Icon was way too big. size got lost with #406


Averethel (approved):

GutierrezGo (approved):
🙈

PR #422, opened 2021-09-16
feat: allow to load image lazy for preview card

talamcol:
Modern browsers support built-in lazy loading for images. It does not break anything in older browsers, it will be just ignored. This adds the ability for this to PreviewCard which is used for garage promotion on the HP

### DIFF
--- a/src/components/previewCard/index.tsx
+++ b/src/components/previewCard/index.tsx
@@ -11,6 +11,7 @@ interface Props {
   cardTitle?: string
   previewLabel: string
   image?: string
+  loading?: "lazy" | "eager"
   logo?: string
   link?: string
   onImageFailedLoading?: (e) => void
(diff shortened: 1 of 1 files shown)


src/components/previewCard/index.tsx
```diff
@@ -11,6 +11,7 @@ interface Props {
   cardTitle?: string
   previewLabel: string
   image?: string
+  loading?: "lazy" | "eager"
```
Averethel, line 14: I think `loading` is a bit too generic when we talk about the whole card. I'd expect the whole component to be lazy-loaded, not just the image. How about:

```suggestion
  imageLoading?: "lazy" | "eager"
```
?

Averethel (approved):
Looks good to me, please check the naming:

src/components/previewCard/index.tsx
talamcol: I agree, will change it, thanks!

PR #423, opened 2021-09-17
feat: added asterisk for required field

djanov:
[CAR-8373](https://autoricardo.atlassian.net/browse/CAR-8373)

adding asterisk for the required field (floatingLabel)

### TODO

- [ ] add requreid for requreid fields in UL 
~~add requreid for requreid fields in DH~~

### DIFF
--- a/src/components/input/index.tsx
+++ b/src/components/input/index.tsx
@@ -59,8 +59,6 @@ interface PopupLabelProps extends InputProps {
 interface FloatingLabelProps extends InputProps {
   labelText?: string
   floatingLabel?: boolean
-  renderLabelPopup?: () => JSX.Element
-  required?: boolean
 }
 
 type Props = PopupLabelProps | FloatingLabelProps
@@ -203,7 +201,7 @@ const Input = forwardRef<HTMLInputElement, Props>(
         name={name}
         error={hasError}
         text={labelText}
-        required={rest.required}
+        required={required}
         {...labelProps}
       >
         {renderInput(hasError)}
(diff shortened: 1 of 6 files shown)


src/components/input/index.tsx
```diff
@@ -59,6 +59,8 @@ interface PopupLabelProps extends InputProps {
 interface FloatingLabelProps extends InputProps {
   labelText?: string
   floatingLabel?: boolean
+  renderLabelPopup?: () => JSX.Element
```
Averethel, line 62: This looks unrelated, why is it needed?

djanov: I got a TS error, that I need to add it on line 102
image

djanov: it's probably because of the ternary nesting, and now we also have the required prop so that's why.

Averethel: It's because of the conditional, you're adding `renderLabalPopup` and `required` if there is `required` in other props. but the presence of `required` does not guarantee the presence of `renderLabelPopup`.

djanov: I just looked at the code, I don't know why I added here anything, I will just remove this part,

Averethel (approved):
Looks good to me. Can you make sure to add the new variant to the storybook before you merge?

djanov:
> Looks good to me. Can you make sure to add the new variant to the storybook before you merge?

I added to the `standard ` in the storybook but sure will make a new one.

PR #427, opened 2021-09-30
feat: add pointers events none to floating label

mreljin:


Averethel (approved):
I'd release it as a `fix`

PR #437, opened 2021-10-12
feat: add electric badge

mreljin:
## Description

https://autoricardo.atlassian.net/browse/CAR-8787

This PR adds the electric badge. Because the electric badge differs in height and background from other badges, it also makes a few changes:
- added `height` prop to base badge which adds extra padding and increases font-size (`size` was reserved for showing text)
- added `background` so a background can be passed (I didn't do this dynamically because it's not good to do so in tailwind)

The other badges are unchanged. Prop values for height and background are hard coded for them so an invalid value can't be passed.

## After

image


src/components/badges/base.tsx
```diff
@@ -2,26 +2,43 @@ import React, { FC, ReactNode } from "react"
...
 
-import { BadgeSize } from "./types"
+import { BadgeBackground, BadgeHeight, BadgeSize } from "./types"
 import styles from "./badges.module.css"
 
 import Tooltip, { TooltipPosition } from "../tooltip/index"
 
 interface Props {
   size: BadgeSize
+  height: BadgeHeight
```
Averethel, line 12: What's the difference between `height` and `size`? From the interface, I have a hard time figuring out which one should I use. Would it be possible to only use one property for sizing?

src/components/badges/types.ts
```diff
@@ -3,6 +3,8 @@ import { ReactNode } from "react"
 import { Language } from "../../types/language"
 
 export type BadgeSize = "small" | "large"
+export type BadgeHeight = "small" | "large"
+export type BadgeBackground = "bg-white" | "bg-whatsapp"
```
Averethel, line 7: How about

```suggestion
export type BadgeBackground = "white" | "green"
```
you can map the property to the concrete colour in the component, no need to do that in the property value. What do you think?

src/components/badges/base.tsx
```diff
@@ -2,26 +2,43 @@ import React, { FC, ReactNode } from "react"
...
       className={classNames(
-        styles.space,
-        "inline-flex items-center border border-grey-1 rounded-4 text-sm font-bold bg-white",
+        height === "small" && styles.space,
+        "inline-flex items-center rounded-4 font-bold",
         {
           "cursor-pointer": tooltipContent,
+        },
+        {
+          "bg-whatsapp": background === "bg-whatsapp",
```
Averethel, line 36: Isn't `whatsapp` a brand colour (the same shade of green as the WhatsApp logo)? Is this really the colour we want to use for the electric badge? Can you clarify that with the designer?

The shades of green that I've found in our palette are:

```
green-light: "#74CC74",
green-opaque: "#E3F4E3",
green-dark: "#149246",
```

Averethel (commented):
A couple of questions and suggestions from my side:

src/components/badges/base.tsx
mreljin: `size` was already used for showing/hiding text in badges. Let's say I only want to show the electric leaf in my badge, I'd set `size="small"`. It was implemented like that before, and it would be a breaking change to rename it. 

So yes, you can use both, or not. Height impacts height only, size impacts whether text is shown.

mreljin: Ah! I thought we'd want to have everything named consistently, and that didn't cross my mind. :+1:

masone: Alternatively, always feel free and try to reduce the amount of variants we have with the designer. when interfaces become too complex to understand, it can be a sign for too complex design requirements

src/components/badges/types.ts
masone: I would push it even further here. The style should be contained with the badge and not passed in from the outside. Is there a way to add the bg color on a wrapper instead of the badge itself?

mreljin: I'd have to add other styles to the wrapper as well, such as rounded corners, borders, set the width of the wrapper... then we'd have to change change the wrapper each time we want something different, instead of sending a prop, which feels weird to me. I feel it's easier to understand and do with it being in one div

also, can you elaborate on the downside of passing it from the outside? This is internal to the components-pkg, and can't be passed from listings and DH

mreljin: > Generally, we aim to be quite on point with the interfaces (ie. props) of shared components in components-pkg.

Agreed! It's worth spending extra time on time, as fixing is later is more time-consuming than getting it right the first time.

> We aim for clear and straight forward interfaces that don't allow any weird combinations that should never happen according to the styleguide.

I took extra care that externally weird combinations cannot be passed as props. So, we cannot have a green money back guarantee badge if it's design is intended to be white. Internally the combinations looked weird because of the naming. I tried to rename it in a fix. Can you take a second look?

> BadgeBackground is currently a css class, which I believe is a violation of proper encapsulation.

I didn't know this! Thanks for the detailed explanation. I was really going off of how somebody corrected me in a review a while back. 

> The badgesize/badgeheight are confusing. I wouldn't really know what effect they have just by looking at the names. I'm also not sure if all combinations would be allowed (small height, large size for example). 

This has been addressed, but the internal mapping to me now looks weird, which is why I suggested we fix it in a follow-up tech ticket anyway. Now externally we pass `size="large"`, but internally `size` maps to `withText`, and to make things more confusing, `size` is also used internally, but in this case it is actually the size (as in width and height).

masone: Ok, if you're all comfy, let's go with this. A lot of this might be preference / style and I don't want to impose too much. Especially if this is an internal component

masone (approved):

PR #455, opened 2021-11-01
BREAKING CHANGE: remove 'SegmentedControl' component

gzimh:
References: [8991](https://autoricardo.atlassian.net/browse/CAR-8991)


lkappeler (commented):
Is this something motivated by the design team as they decided that they won't use segmented controls in the future anymore? My thinking behind is even if we are currently not using a segmented control we might still wanna keep it in case a new/iterated feature wants to make use of the component.

gzimh:
@lkappeler the way i understand is that they don't want to use these 'tabs' anymore. One of the tickets of listing reactivation topic was to change the looks of the tabs because of the 'outdated' look these had.

masone:
I have the same understanding as Gzim. I remember Enrique said the element was generally misunderstood by users. Always fair to double check the designers intention though. On the other hand - should it ever come up again, it's a button click away from being reverted

lkappeler (approved):
Thanks for the explanations 👍

masone:
in case this did not get released, I believe it's because as far as I understand, the commit message still needs a fix/chore/feat and the breaking change indicator has to be in the commit message body, for example:

```
chore: remove SegmentedControl component

BREAKING CHANGE: `SegmentedControl` is unused and was removed. use xyz instead
```

gzimh:
@masone It looks it didnt get released.. what do you recommend now since i already merged this PR?

masone:
Unfortunately, an empty commit is the only way in this situation (we don't want to alter history on master). 

`git commit -am "your full message" --allow-empty`

We'll loose the reference from the commit to the PR that includes the breaking change and the changelog will be a bit wonky. But it happens. Let's make sure it's properly released and we should be fine

Please do so timely - ie. before anyone else releases some fix that would then include this breaking commit as well :)

PR #458, opened 2021-11-02
chore: remove SegmentedControl component

gzimh:
BREAKING CHANGE: 'SegmentedControl' is unused and was removed.

REF: https://github.com/carforyou/carforyou-components-pkg/pull/455


masone (approved):
commit message looking good

PR #460, opened 2021-11-03
fix: CLS above the fold for the new range filter

talamcol:
CSS modules kick in too late. Since the new filter for price is above the fold, this results in a layout shift. Using global CSS with tailwind hopefully solves this issue.


talamcol:
created a tech task to investigate if this behavior can be changed: https://autoricardo.atlassian.net/browse/CAR-9076

Marine-Berthier (approved):

PR #477, opened 2021-11-09
CAR-8998 flash messages

lkappeler:
https://autoricardo.atlassian.net/browse/CAR-8998

Exports the flash message concept from dealerhub and unifies the style with our alert messages.

Screenshot 2021-11-09 at 18 04 11
Screenshot 2021-11-09 at 18 03 57

In order to provide flash messages in the application, the provider and container must be added to the app
```
<FlashMessagesProvider>
          <FlashMessages />
          <p>Content</p>
</FlashMessagesProvider>
````

To use add a flash message from any component the provided context can be used
you must specify a type and the content. optionally a delay in ms can be defined to automatically close the message. further on an optional onClose handler can be passed to the message if neeed.
```
  const { addMessage } = useContext(FlashMessagesContext)

  const add = () => addMessage({ type, content: dummyText, autoCloseDelay })
```

### DIFF
--- a/src/components/alertMessage.tsx
+++ b/src/components/alertMessage.tsx
@@ -1,6 +1,7 @@
 import React, { FC, ReactNode } from "react"
 import classnames from "classnames"
 
+export type AlertMessageType = "error" | "warning" | "information" | "success"
 interface Props {
   /**
    * Any element to be displayed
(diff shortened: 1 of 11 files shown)


masone (approved):
looking 🔥 to me

PR #478, opened 2021-11-10
Car 9104: Slider

Marine-Berthier:
New component for slider with input : [CAR-9104](https://autoricardo.atlassian.net/browse/CAR-9104?atlOrigin=eyJpIjoiODNiNGUzYzRiOTEzNGQ0NmFjNmE2NjUyZDM2NmI1YmYiLCJwIjoiaiJ9)

[PR in listings](https://github.com/carforyou/carforyou-listings-web/pull/3781) 

slider

Here is an example in listings : https://slider-9104.carforyou-listings-web.branch.carforyou.ch/en/loan/882224/apply


src/components/input/index.tsx
```diff
@@ -141,18 +141,22 @@ const Input = forwardRef<HTMLInputElement, Props>(
...
-          "floatingLabel-input": labelProps.floating,
-          "input_left hover:z-1 hover:transition hover:duration-200 focus:z-1":
-            position === "left",
-          "input_right -ml-1 hover:z-1 hover:transition hover:duration-200 focus:z-1":
-            position === "right",
-          "text-right": textAlignment === "right",
-        })}
+        className={classNames(
+          "w-12/12",
+          textAlignment === "center-right" ? "pl-45 pr-10 pb-3" : "px-10 py-13",
```
mreljin, line 146: This `px-10 py-13` is deleted from the base input above, correct? Is there no way to overwrite it here instead of deleting it from the base input? 

also it seems a bit weird, it seems as if the condition can be moved inside the brackets `{}`. We have this condition above the brackets, while `"text-right": textAlignment === "right",` is inside

mreljin: also `center-right` sounds a bit weird as well, from looking at it I'm not sure what it does

src/tailwind/defaultConfig.ts
```diff
@@ -424,6 +424,7 @@ export default {
       "25": "25px",
       "30": "30px",
       "40": "40px",
+      "45": "45px",
```
Marine-Berthier, line 427: TO DO : Remove

src/components/input/index.tsx
Marine-Berthier: I wanted to stick to the design but overcomplicated the task 👎 
Checked with Enrique who is fine having the unit as part of the label, so I removed this part 👍

src/components/slider.tsx
```diff
@@ -0,0 +1,88 @@
...
+  step,
+  min,
+  max,
+  name,
+  defaultValue,
+  inputLabel,
+  required = false,
+}) => {
+  const [rangeValues, setRangeValues] = useState([defaultValue])
+  const [inputValue, setInputValue] = useState(defaultValue)
```
talamcol, line 28: Is it maybe possible to remove this state variable and use `String(rangeValues[0])` instead for the input field?

src/components/slider.tsx
```diff
@@ -0,0 +1,88 @@
...
+        />
+        <InputFilter
+          name={name}
+          initialValue={String(inputValue)}
+          mode="numeric"
+          apply={(e) => {
+            e.target.value > max
+              ? setRangeValues([max])
+              : setRangeValues([e.target.value])
+            e.target.value === "" && setRangeValues([min])
```
talamcol, line 76: does that mean the slider is no longer in sync with the input field here?

src/components/slider.tsx
```diff
@@ -0,0 +1,88 @@
...
+            setInputValue(e.target.value)
+          }}
+          textAlignment="left"
```
talamcol, line 79: I think this is the default and you can remove it

talamcol (approved):
looking great! Maybe ask someone independent to review it too since I'm a bit biased :D and let's wait with the merge until we properly integrated it into the form in case we missed something

edit: I approved by mistake, sorry :)

src/components/slider.tsx
```diff
@@ -0,0 +1,88 @@
...
+import colors from "../tailwind/colors"
+
+interface Props {
+  step: number
+  min: number
+  max: number
+  name: string
+  defaultValue: number
+  inputLabel: string
+  required?: boolean
```
talamcol, line 15: we need an `onChange` handler here because otherwise we cannot control it with formik

src/components/slider.tsx
Marine-Berthier: Yes, kind of , but it's only in case the input is empty. Without this condition, when input is empty the cursor is no more on the slider

mreljin (approved):
Looking good after the recent changes :+1:

src/components/slider.tsx
```diff
@@ -0,0 +1,91 @@
...
+  handleChange?: (value: number) => void
```
talamcol, line 16: should not be optional

src/components/slider.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <>
+      <Label fieldName={inputLabel} required={required} />
+      <div className="flex flex-col-reverse md:flex-col md:pt-15 md:space-y-20">
+        <Range
+          values={rangeValues}
+          step={step}
+          min={min}
+          max={max}
+          onChange={(values) => {
+            handleChange(values[0])
```
talamcol, line 42: please only call this for `onFinalRelease`

talamcol (changes requested):
requesting changes since you cannot remove the approval that I accidentally added

talamcol (approved):

lkappeler (commented):
I like it so far, played a bit around in the UI and found some things we potentially wanna handle

Not sure if this is intended or not, the circle from the slider is wider than the input on the min/max values. We might wanna add some padding so the border of the input is the widest part as otherwise we might end up with the input being smaller than a different one without a slider
image

When I enter for example 2000 it resets it to 1000 but when I enter it again it stays

https://user-images.githubusercontent.com/3371760/141293738-0aefd8a0-d308-4d34-99f4-c4fd3a6bb52f.mov

When using the keyboard to fill the input I can enter negative values which leads to a funny state in the slider


https://user-images.githubusercontent.com/3371760/141293818-4fd0e880-ce81-49df-8dab-f5f48599d33c.mov

Marine-Berthier:
Thank you for your review Lorenzo ! 
I added some padding, I think it looks weird to be honest because the slider's width is smaller than the input's width 
padding slider
It is no more possible to enter negative values, however I am still looking for a way to fix the max value reset (well spotted by the way 👍)

lkappeler:
@Marine-Berthier Thanks for the fixes. Regarding the pairing, you are right it kinda looks strange in both versions not quite sure what's best here. Maybe you can check with a designer if they had an intention when designing the slider or rather the range inputs. I took a look on how they are in production and it somewhat inconsistent already
Mobile (advanced filters)
Screenshot 2021-11-15 at 12 52 03
Desktop (Sidebar)
Screenshot 2021-11-15 at 12 50 50

Marine-Berthier:
Thank you @Lorenzo ! 
I'll check with Enrique, if needed I'll create a ticket to  make the sliders consistent.
About the edge cases for `max` and `min`, I checked with Colin and we could not find a nice way to handle them. In this case we take advantage of the BE validation to return an error message and prevent the value to be sent if it's not between the `min` and `max`. We don't force the value but we let the user know that it's wrong.

talamcol:
> @Marine-Berthier Thanks for the fixes. Regarding the pairing, you are right it kinda looks strange in both versions not quite sure what's best here. Maybe you can check with a designer if they had an intention when designing the slider or rather the range inputs. I took a look on how they are in production and it somewhat inconsistent already Mobile (advanced filters) Screenshot 2021-11-15 at 12 52 03 Desktop (Sidebar) Screenshot 2021-11-15 at 12 50 50

this is by design. Enrique wanted larger padding on mobile because sometimes the device gestures are triggered when sliding near the screen limit. The second screenshot is desktop, that's why it has no padding

PR #481, opened 2021-11-11
fix: restructure checkbox component

Marine-Berthier:
[CAR-9080](https://autoricardo.atlassian.net/browse/CAR-9080?atlOrigin=eyJpIjoiNjI4NTgyZjMxYzkyNDBjNWJjMzhjNWIyMDA2YzgyOWUiLCJwIjoiaiJ9)

`div` is not allowed to be a child of `label` as it’s a block element (invalid HTML). Also, the `input` element should not be a child of `label` but a sibling. We should restructure it to have valid HTML.

You can check that Checkoxes and radio buttons are still working fine here : https://github.com/carforyou/carforyou-dealerhub-web/pull/1681


src/components/radioButton.tsx
```diff
@@ -46,14 +46,16 @@ function RadioButton({
     <label
```
Marine-Berthier, line 46: I am trying to remove this wrapping label but if I replace it by a div for example the radio input is no more clickable, only the label is. If someone has an idea why, please let me know 👍

talamcol: what happens if you remove the htmlFor as well?

talamcol (commented):
thank you! we still have 5 levels of nested div's that can definitely be simplified but at least or HTML is now valid. Looks good to me, maybe ask QA for a quick test on checkboxes and radio buttons

src/components/radioButton.tsx
Marine-Berthier: Same, the input is still not clickable

talamcol: I quickly checked and for me it's all working fine. The issue that I can see that in the radioButton.stories.tsx there is no `name` passed as a prop. As this name is going to be the ID of the radio button and the `htmlFor` of the label, it is mandatory. If I pass name in stories, it works as expected.

Marine-Berthier: As you suggested the issue was due to CSS. We don't actually use the native input but a custom one. I gave the native input a width and a height and I set a z-index for it to be triggered on click.

src/components/radioButton.tsx
```diff
@@ -43,17 +43,18 @@ function RadioButton({
...
+        "w-12/12",
+        disabled ? "cursor-not-allowed text-grey-4" : "cursor-pointer"
       )}
-      htmlFor={name}
     >
       <WithValidationError error={error}>
         {(hasError) =>
           renderLabel ? (
             <WithHorizontalLabel
+              name={name}
```
masone, line 56: on naming: it would not be immediately obvious to me that the name I'm passing must match the name of the input. as this is a "label" component, maybe just stick to `htmlFor`, which would maybe allow better to infer the purpose?

it's probably a nitpicky comment. if we only use the `HorizontalLabel` internally, the naming is fine. if we export it, I'd maybe consider giving it another thought

src/stories/radioButton.stories.tsx
```diff
@@ -13,6 +13,7 @@ export default {
   title: "Radio Button",
   component: RadioButton,
   args: {
+    name: "radioButton",
```
masone, line 16: you didn't change anything to the props - how did this not throw a type error? shouldn't the Prop types have enforced passing a name here?

src/components/fieldHelpers/withHorizontalLabel.tsx
```diff
@@ -32,7 +36,15 @@ const WithHorizontalLabel: FC<Props> = ({
...
-    <span className="w-12/12">{renderContent && renderContent()}</span>
+    <label
+      htmlFor={name}
+      className={classNames("w-12/12", {
+        "cursor-not-allowed text-grey-4": disabled,
+        "cursor-pointer": !disabled,
+      })}
+    >
+      {renderContent && renderContent()}
+    </label>
```
masone, line 47: do we really want to render an empty `<label></label>` when `renderContent` is not passed?

masone: alternatively, make `renderContent` a required prop (we might use it like this anyway already) and drop the conditional

masone (commented):
nice restructuring! minor questions/comments from my side 😃

assets/input/radio.css
```diff
@@ -3,9 +3,8 @@
 }
 
 .inputRadio-input {
-  @apply absolute opacity-0 cursor-pointer;
-  height: 0;
-  width: 0;
+  @apply absolute opacity-0 cursor-pointer h-checkbox w-checkbox;
```
talamcol, line 6: still a bit surprised that it was the CSS causing it not to work 🤣

talamcol (approved):
we are down to 68 HTML errors on the advanced search from an initial **206** errors (SRP it's from 194 to 73 errors). 
Fantastic! Plus one for the comments from Christian, otherwise it looks good to me. I tested listing a bit and it all seems to work as before. I would assign QA nevertheless though.

src/components/radioButton.tsx
Marine-Berthier: Thank you for the review. 🙏 
We use the label component internally currently but I changed `name` for `htmlFor`, it's more meaningful!

src/stories/radioButton.stories.tsx
Marine-Berthier: I tried with other components such as AlertMessage. I removed the `type` (which is a mandatory prop) from the stories and it does not complain 🤯

masone: Weird! Let's flag this mentally - if we don't figure it out on the go, we can follow up with a proper task

PR #508, opened 2021-11-30
fix: added missing tailwind value

GutierrezGo:
Reference: https://github.com/carforyou/carforyou-dealerhub-web/pull/1707

## Motivation
While I was exporting the flash message component I noticed that we were missing the z-index value.

# Thank you 🐥


lkappeler (approved):
Thanks 🙂

PR #509, opened 2021-12-01
fix: fixed horizontal scrollbar

GutierrezGo:
Reference: https://autoricardo.atlassian.net/browse/CAR-9156

## Take into consideration
The scrollbar issue is reproduced ONLY on Firefox.

## Before
Screenshot 2021-12-01 at 15 17 29

## After

Screenshot 2021-12-01 at 15 19 29

# Thank you ☃️


ufkaya (approved):

masone:
Thanks, that was annoying :D

PR #510, opened 2021-12-03
fix: removing z-index for clear icon

djanov:
Reference: [CAR-9313](https://autoricardo.atlassian.net/browse/CAR-9313)

We have a bug with the clear icon in the input field because of the **z-index**, now it's removed. I have a [follow-up ticket for UL](https://github.com/carforyou/carforyou-listings-web/pull/3872) for **QA** to check.

[cleanup for DH](https://github.com/carforyou/carforyou-dealerhub-web/pull/1722)

### DIFF
--- a/src/__tests__/__snapshots__/node.test.ts.snap
+++ b/src/__tests__/__snapshots__/node.test.ts.snap
@@ -422,7 +422,6 @@ Object {
       "1": 1,
       "auto": "auto",
       "card": 10,
-      "clearButton": 10,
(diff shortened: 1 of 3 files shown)


masone (approved):
z-indexes are hard and I usually feel uncomfy changing them without understanding why it was introduced in the first place. it must have fixed something back then... code looks straight forward though. give it a thorough test wherever it's used. test modals, overlaying dropdowns, intercom buttons, floating buttons, etc...

PR #515, opened 2021-12-06
fix: undefined slider tracking

talamcol:
References: [CAR-9318](https://autoricardo.atlassian.net/browse/CAR-9318)

If the user used the keyboard (arrowup, down) or clicked on the slider to change the value of the thumb, the tracking resulted in `undefinedSlider`. Furthermore, the tracking and addFilter event also run if the slider value didn't change.


Marine-Berthier (approved):
Thank you Colin !

PR #516, opened 2021-12-06
feat: allow to pass a placeholder to the range filter

talamcol:
References: [CAR-9093](https://autoricardo.atlassian.net/browse/CAR-9093)

The slider should be able to show a placeholder to help the user understand possible values. Furthermore, the year field does not have a unit thus it needs to get changed to optional. 
You can see how it looks here: https://github.com/carforyou/carforyou-listings-web/pull/3840

### DIFF
--- a/src/stories/filters/rangeWithFacets.stories.tsx
+++ b/src/stories/filters/rangeWithFacets.stories.tsx
@@ -151,10 +151,7 @@ WithoutUnit.args = {
     min: "priceFrom",
     max: "priceTo",
   },
-  inputPlaceholder: {
-    min: "0",
-    max: "30000+",
-  },
+
   value: {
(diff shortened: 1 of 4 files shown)


src/stories/filters/rangeWithFacets.stories.tsx
```diff
@@ -122,10 +130,31 @@ WithoutFacets.args = {
...
+    min: "priceFrom",
+    max: "priceTo",
+  },
+
+  value: {
+    min: null,
+    max: null,
+  },
+  subtext: "(1450 Autos)",
+}
```
masone, line 160: 👍 thanks for adding these new variants to storybook

masone (approved):

PR #525, opened 2021-12-15
fix(deps): update storybook, require React 17

masone:
It's still impossible to `npm link` our components-pkg in listings without running into the infamous "invalid react hooks error". Our current leads point towards ending up with two React versions in the project, despite them being dev dependencies here. Updating storybook allows us to use v17 - it's worth a shot

BREAKING CHANGE: Requires React 17

can't yet test it locally because only 1 package can be linked at the time. I'd need both components and header-footer to verify 🙈 it's a good step nonetheless


package.json
```diff
@@ -83,9 +78,7 @@
     "semantic-release": "17.4.7",
     "tailwindcss": "2.2.19",
     "ts-jest": "27.0.7",
-    "ts-loader": "8.3.0",
-    "typescript": "4.3.5",
-    "webpack": "4.46.0"
```
masone, line 88: a lot of dependencies are not needed. they're likely implicit with the new storybook version

package.json
```diff
@@ -38,43 +39,37 @@
...
     "@types/react": "16.14.21",
     "@types/react-paginate": "7.1.1",
     "autoprefixer": "10.4.0",
-    "babel-loader": "8.2.3",
     "bundlesize": "0.18.1",
     "eslint": "8.4.1",
-    "eslint-plugin-react-hooks": "4.3.0",
     "jest": "27.3.1",
-    "marked": "2.1.3",
-    "marked-terminal": "4.1.1",
```
masone, line 77: I believe those were added as workarounds for something. we don't use them anywhere explicitly

package.json
```diff
@@ -46,46 +46,37 @@
...
     "@carforyou/assets": "2.1.2",
     "@carforyou/eslint-config": "4.0.0",
     "@carforyou/pika-plugin-build-components": "1.2.4",
     "@pika/pack": "0.5.0",
     "@pika/plugin-copy-assets": "0.9.2",
     "@pika/plugin-standard-pkg": "0.9.2",
     "@pika/plugin-ts-standard-pkg": "0.9.2",
-    "@storybook/addon-actions": "6.3.8",
-    "@storybook/addon-essentials": "6.3.8",
-    "@storybook/addon-links": "6.3.8",
```
masone, line 59: these are all "essentials"...

package.json
```diff
@@ -25,8 +25,8 @@
   "license": "MIT",
   "homepage": "https://github.com/carforyou/carforyou-components-pkg#readme",
   "peerDependencies": {
-    "react": ">=16.12.0",
-    "react-dom": ">=16.12.0",
+    "react": ">=17.0.0",
+    "react-dom": ">=17.0.0",
```
Averethel, line 29: Same as in header-footer, maybe we can update dev dependencies only?

package.json
```diff
@@ -46,46 +46,37 @@
...
     "@types/jest": "27.0.3",
-    "@types/node": "14.18.0",
+    "@types/node": "16.11.13",
     "@types/react": "16.14.21",
     "@types/react-paginate": "7.1.1",
     "autoprefixer": "10.4.0",
-    "babel-loader": "8.2.3",
     "bundlesize": "0.18.1",
     "eslint": "8.4.1",
-    "eslint-plugin-react-hooks": "4.3.0",
```
Averethel, line 74: Don't we use this?

Averethel (approved):

package.json
masone: I would expect this to be in estlint-config-pkg. linter works fine and removing makes no difference

package.json
masone: yes-ish. it gives me extra confidence that there would be no react16 laying around anywhere in other deps as it would halt the install. as we're on react17 already everywhere, I felt it's a safe move. I'm happy to adjust though. let me know what you think...

package.json
Averethel: Oh, it's moved there indeed. I recall seeing a ticket about it so I wasn't aware it was done. (it's also possible that we're not using it at all here)

package.json
Averethel: Okay, I see. I'm not 100% sure it would halt the installation since react 17 allows multiple versions to co-exist IIRC.
I'm fine with going with it as is, no strong opinion.

masone: fair point, but our packages would now prevent us from having an implicit v16 somewhere. I'll move on with merging if your opinion is not too strong. I can always follow up if it causes issues

masone:
I could verify it with a prerelease and a linked package. local linking now works in listings!!!!

PR #526, opened 2021-12-16
chore: remove npm link workarounds 

masone:
with #525, no more workarounds are needed to link the package. official workflows work just as expected. this removes any custom recommendations


Averethel (approved):
GONE

PR #536, opened 2021-12-27
fix: building components with rollup (BREAKING CHANGE)

masone:
Todo: 
- [x] issue a breaking change
- [x] test prerelease

References carforyou/carforyou-docs#147

Moving away from pika/pack, following the decisions outlined in the RFC above. Zero dependencies for a build

- removed pika/pack
- moved stories to .storybook for clearer separation
- removed node entrypoint. it's not necessary with proper treeshaking


lkappeler (approved):
Nice job looks good to me

bundlesize.config.json
```diff
@@ -1,8 +1,8 @@
 {
   "files": [
     {
-      "path": "pkg/dist-web/index.js",
-      "maxSize": "25 kB"
+      "path": "pkg/cjs/index.js",
+      "maxSize": "30 kB"
```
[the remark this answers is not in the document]
masone: the check was off - `dist-web` was building separate files since #459 and the check was only focussing on the entrypoint instead of the full bundle

.storybook/main.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  stories: ['../src/**/*.stories.@(js|tsx)'],
+  stories: ['./stories/**/*.stories.@(js|tsx)'],
```
Averethel, line 2: Why are we piggybacking on collocating stories with components? We've put an effort into it to have it well grouped and improve developer experience in that regard. Interested to hear your motivation here.

.storybook/stories.css
```diff
@@ -1,4 +1,5 @@
 @tailwind base;
+@import "../assets/index.css";
```
Averethel, line 2: Shouldn't the custom CSS come after all tailwind?

.storybook/main.js
Averethel: Oh nvm, they weren't truly collocated, but in `src/stories` instead...

.storybook/stories/alertMessage.stories.tsx
```diff
@@ -2,8 +2,8 @@ import React, { FC } from "react"
 
 import { StoryProps } from "./storyProps"
 import StoryContainer from "./storyContainer"
-import AlertMessage, { AlertMessageProps } from "../components/alertMessage"
-import MailSent from "../../.storybook/icons/mailSent"
+import AlertMessage, { AlertMessageProps } from "../../src/components/alertMessage"
```
Averethel, line 5: Food for thought - if we import from the index we'd have a bit less pain moving those around. Plus it would fail if the component isn't exported from the package.

rollup.config.js
```diff
@@ -0,0 +1,49 @@
...
+        preserveModules: true,
+        preserveModulesRoot: "src",
+      },
+    ],
+    plugins: [
+      resolve(),
+      commonjs(),
+      typescript({
+        tsconfig: "./tsconfig.json",
+        exclude: ["**/__*__/**/*"],
```
Averethel, line 36: Don't we need to exclude stories too?

src/components/badges/__tests__/buyNow.test.tsx
```diff
@@ -2,7 +2,7 @@ import React from "react"
 import { fireEvent, render, screen } from "@testing-library/react"
 
 import BuyNowBadge from "../buyNow"
-import { DummyTooltip } from "../../../stories/badges/dummyTooltip"
+import { DummyTooltip } from "../../../../.storybook/stories/badges/dummyTooltip"
```
Averethel, line 5: I would've not expected to import from stories in tests, looks very weird. 
Can we extract those `dummy` components somewhere where they are properly shared? 
(not necessarily now, but in a follow-up tackled in due time)

src/components/filters/input.tsx
```diff
@@ -5,7 +5,7 @@ import Input from "../input/index"
 import { scrollIntoViewIfMobile } from "../../lib/scrollHelper"
 import getClosestElement from "../../lib/getClosestElement"
 
-interface Props {
+export interface Props {
```
Averethel, line 8: What's the motivation for exporting props?

Averethel (approved):
Looks good to me, I have a few questions, nothing blocking though.

.storybook/main.js
masone: yes, I had the exactly same thought process - moving them all to src was not was we intended. since storybook and the build need different configs sometimes, I ended up separating completely

.storybook/stories.css
masone: no, according to the docs this is where they're supposed to be and it's in line with what we have in listings. dind't re-check, speaking from memory

.storybook/stories/alertMessage.stories.tsx
masone: very good point - I'll move that to another effort though

rollup.config.js
masone: not now that they're in `.storybook` - that's one of the simplifications I was able to make with the move

src/components/badges/__tests__/buyNow.test.tsx
masone: absolutely - had the same reaction. I already touched so many things in this PR and would like to push that to a separate effort

src/components/filters/input.tsx
masone: thanks for asking - not obvious at all. the updated storybook dependencies expect an interface Props to be exported (I guess to generate prop docs?) - I don't recall what tooling, but something was yelling at me about it

.storybook/main.js
Averethel: Do we still want to properly co-locate them? What is your view on it?

src/components/filters/input.tsx
Averethel: Okay, that makes sense. I suppose that might be then tricky to integrate with the index export, but worth trying still.

PR #537, opened 2021-12-27
chore: remove custom plugins

masone:
BREAKING CHANGE: removed custom gradients plugin. use native functionality instead https://v2.tailwindcss.com/docs/gradient-color-stops
BREAKING CHANGE: removed custom rotate plugin. use native functionality instead https://v2.tailwindcss.com/docs/rotate

`gradients` forces a full `lodash` dependency into our client-side bundle apart from another large dependency https://github.com/benface/tailwindcss-gradients/blob/master/package.json. this was the reason we introduced a separate node entrypoint. while we're cleaning up, let's also remove the rotation one. both have equivalent core plugins by now


masone:
superseeded by https://github.com/carforyou/carforyou-components-pkg/pull/538

PR #538, opened 2021-12-28
Exploration: 3 potential improvements

masone:
In the course of migrating to ESM (https://github.com/carforyou/carforyou-docs/discussions/147), I went down the rabbit hole and reviewed some of out past decisions. This PR is a mishmash of experiments I ran while tinkering around. Nothing is strictly necessary, but we can discuss what we want to push:

## tailwind config
remove the build step for a static tailwind config. we can just require the src files if we play it right

## remove plugins

If you remember, we at one point ended up with a full `lodash` dependency in our client-side bundle. It [originates in `tailwindcss-gradients`](https://github.com/benface/tailwindcss-gradients/blob/master/package.json). We solved this by separating node/browser with two entrypoints, which is the reason for our custom [pika plugin](https://github.com/carforyou/pika-plugin-build-components-pkg). This was a decision was probably the wrong call in the first place (just because someone requires a commonjs module, it doesn't necessarily mean that they're on node) It added a lot of complexity to our setup. With proper treeshaking we can revert that decision - lodash does not end up in the client bundle. I discovered that the two custom plugins are now covered by the core and we can remove them. 

- remove custom gradients plugin. this is built into [the core by now](https://v2.tailwindcss.com/docs/gradient-color-stops). this removes `lodash` from our bundle
- remove custom rotate plugin. this too, is [natively available](https://v2.tailwindcss.com/docs/rotate)

## expose theme instead of config
only keep the tailwind theme in `components-pkg` and delegate the ownership of the full config to the projects. I exposed the theme instead of the config to prevent external dependencies leaking in. I only later realized, that this does not apply anymore in our current treeshakeable package. The last point is not necessary, but we can still discuss if we want it anyway. Should we ever want to leverage custom plugins again, they would have to be duplicated across projects. My gut feeling here is, that we probably want to encourage rolling with the defaults. I think this is safe - we've made it up to here with very little customization after all. Because this package is primarily used client-side, I suggest we discourage the introduction of any server-side-specific code entirely. This opens the possibility to consume the theme client-side (it might still be too large, but at least no dependencies are included anymore). 

The configuration in the projects stays pretty close to what it is today. Apart from the theme, there was nothing really substantial which was shared and most of the time, we fully rely on tailwinds defaults. However, should we ever want to go back to exposing a full config, we'd have to issue another breaking change.

--- 

For the sake of documentation, these are the classes generated by the two plugins - we don't use them heavily at all

```
.bg-gradient-to-bottom-grey{
  background-image:linear-gradient(#f9f9fb,#fff 20%)
}
.bg-gradient-to-bottom-black{
  background-image:linear-gradient(transparent,transparent,rgba(0,0,0,.5))
}

.customRotate-90{
  transform:rotate(90deg)
}
.customRotate-180{
  transform:rotate(180deg)
}
.customRotate-270{
  transform:rotate(270deg)
}
```


masone:
Decision: 
- created tickets for removing the plugins
- will remove the build step for tailwind config
- will keep exposing the full config. lilttle upside, big pain in changing. let's wait

PR #542, opened 2022-01-03
chore: adjusted resources value

GutierrezGo:
Datadog dashboard:
https://app.datadoghq.com/notebook/1750667/virginia-jan-04-2022-13-56?start=1640705304684&range=604800000&live=true

## Take into consideration
The branch deployments are on the `teal` cluster.

### DIFF
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -20,8 +20,8 @@ ingress:
 
 resources:
   limits:
-    cpu: 500m
-    memory: 500M
-  requests:
-    cpu: 250m
+    cpu: 30m
     memory: 200M
+  requests:
+    cpu: 10m
+    memory: 100M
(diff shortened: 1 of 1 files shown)


Averethel (approved):

GutierrezGo:
@masone as you said, it's almost the same as  the components.  Thank you for the feedback, time and inputs.
Screenshot 2022-01-04 at 17 08 07

masone (approved):
Thanks Virginia!

For the sake of documenting our call:
 - we initially looked at graphs with _all_ branch deployments. we have to look at the components-pkg-specific one instead. the numbers were of course much higher if we look at the avg/max of all branch deployments as they also include "hungry" apps like listings
 - looking at the max gives us the right view, even if the avg is below the threshold, the pod might still have been restarted when it peaks for some seconds above the limit

PR #546, opened 2022-01-04
chore: add viewport addon

GutierrezGo:
## Motivation
While I was checking a branch I found that the Storybook Viewport Addon was removed.

Storybook Viewport Addon -> https://storybook.js.org/docs/react/essentials/viewport

I used a lot that addon when I'm working on Storybook. And now that we have the branch deployments on components and header-footer I think, it will be a good practice to use this nice addon that Story provide us.

Maybe you are thinking -> but we can resize the window!. Yes, you can. BUT remember that you are resizing the Storybook itself as well and no ONLY the component.

## Take into consideration
I added a configuration for the breakpoints.

Screenshot 2022-01-04 at 17 53 03

# Thank you 🛎

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,2 +1,29 @@
-import "./stories.css"
-import "../assets/index.css"
+require("../assets/index.css")
+require("./stories.css")
+
+import { addParameters } from '@storybook/react';
+
+const screens = {
+  xxs: '375px',
+  xs: '480px',
+  sm: '576px',
+  md: '768px',
+  lg: '1024px',
+  xl: '1280px',
+  xxl: '1920px'
+}
+
+const customViewports = Object.entries(screens).reduce((acc, [key, value]) => {
+  acc[key] = {
+    name: key,
+    styles: {
+      width: value,
+      height: "100%"
+    }
+  }
+  return acc
+}, {})
+
+addParameters({
+  viewport: { viewports: customViewports },
+})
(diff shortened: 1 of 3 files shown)


Marine-Berthier (approved):
Thank you Virginia !

Averethel (approved):

.storybook/preview.js
```diff
@@ -1,2 +1,29 @@
...
+
+const screens = {
+  xxs: '375px',
+  xs: '480px',
+  sm: '576px',
+  md: '768px',
+  lg: '1024px',
+  xl: '1280px',
+  xxl: '1920px'
+}
```
masone, line 14: is it possible to require the `defaultConfig` and iterate over `tailwindConfig.theme.screens` instead of recreating the object here?

GutierrezGo: Ohhh nice idea. Cleaning, cleaning 🧼 🧹 🧽

ufkaya: If I remember correctly, by importing the tailwind config, we will increase the bundle size, that's why when we first added it, we also hardcoded the screen sizes

masone: correct, but this should be local to storybook and I would not care about its bundle size. if it's too much of a hassle, let's duplicate it. in the course of the esm build, I've also been thinking about just exposing the theme and not the full config, so no third parties can sneak in here. most of them are unused by now anyway

masone (approved):

PR #549, opened 2022-01-06
fix: inputFilter

Marine-Berthier:
References [CAR-9423](https://autoricardo.atlassian.net/browse/CAR-9423?atlOrigin=eyJpIjoiOTYxYjE1MDVkYTY1NGYxZWI1ODNkNTFlYjcyZjg4MWQiLCJwIjoiaiJ9)

Related PR : [DH](https://github.com/carforyou/carforyou-dealerhub-web/pull/1782), [Listings](https://github.com/carforyou/carforyou-listings-web/pull/3961)

## Before
Type field is automatically filled with “undefined”

## After
Type field is left empty

### DIFF
--- a/src/components/filters/input.tsx
+++ b/src/components/filters/input.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useEffect, useRef, useState } from "react"
+import React, { FC, useRef, useState } from "react"
 
 import Input from "../input/index"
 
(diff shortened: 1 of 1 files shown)


src/components/filters/input.tsx
```diff
@@ -40,7 +40,8 @@ const InputFilter: FC<Props> = ({
   const inputRef = useRef()
 
   useEffect(() => {
-    ;(inputRef.current as HTMLInputElement).value = initialValue
+    ;(inputRef.current as HTMLInputElement).value =
+      initialValue === undefined ? "" : initialValue
```
ufkaya, line 44: if we want to also cover in case the value is `null`, we can use the nullish coalescing operator `??`
`;(inputRef.current as HTMLInputElement).value = initialValue ?? ""`

ufkaya (approved):
Thank you for the fix👍

src/components/filters/input.tsx
Marine-Berthier: Yes, thanks a lot for the suggestion !

lkappeler (approved):

PR #553, opened 2022-01-11
fix: removed duplicated values

GutierrezGo:
References: https://autoricardo.atlassian.net/browse/CAR-8834

## Motivation and context

Before I start working on the tailwind upgrade, I'm cleaning up and killing some tech20 tickets 🪓

## Take into consideration
I removed the values that we don't use on listings, DH and header-footer and we duplicate them ONLY on the default config.

## How to test
The UI stays unchanged.

# Thank you 🔫


src/__tests__/__snapshots__/node.test.ts.snap
```diff
@@ -319,7 +316,6 @@ Object {
       "40": "40px",
       "5": "5px",
       "8": "8px",
-      "floatingLabelInputFocused": "12px",
```
masone, line 322: plus one. these don't make much sense as "named" ones. we started out naming magic numbers. by now a lot of them are just regular values (we have way too many of them anyway). plus we have css modules for true one-offs. good call!

masone (approved):
clearer is better. please give it a quick 👀 in listings before merging

PR #573, opened 2022-01-25
Car 9534: new component "radioButtonGroupWithTitle"

Marine-Berthier:
References [CAR-9534](https://autoricardo.atlassian.net/browse/CAR-9534)
Related PR : https://github.com/carforyou/carforyou-listings-web/pull/4030

Currently we dont have a component that returns a group of `RadioButton` with a global title. It could be useful in some cases, for example, to solve the issue of the related PR linked above.

In case this PR is approved, I'll create a task to replace the current occurences

### DIFF
--- a/src/components/radioButtonGroupWithTitle.tsx
+++ b/src/components/radioButtonGroupWithTitle.tsx
@@ -6,18 +6,20 @@ import Label from "./fieldHelpers/label"
 interface Props {
   title: string
   required?: boolean
+  disabled?: boolean
   radioInputs: RadioButtonProps[]
 }
 
 function RadioButtonGroup({
   title,
   required = false,
+  disabled = false,
   radioInputs,
 }: Props): ReactElement {
   return (
     <div className="w-12/12">
       <Label fieldName={title} required={required} />
-      <div className="flex w-12/12 justify-between">
+      <div className="flex justify-between">
         {radioInputs.map((radioInput, index) => (
           <RadioButton
             key={index}
@@ -26,7 +28,7 @@ function RadioButtonGroup({
             value={radioInput.value}
             checked={radioInput.checked}
             renderLabel={radioInput.renderLabel}
-            labelPosition={radioInput.labelPosition}
+            disabled={disabled}
           />
         ))}
       </div>
--- a/src/components/radioButtonGroupWithTitle.tsx
+++ b/src/components/radioButtonGroupWithTitle.tsx
@@ -1,6 +1,7 @@
-import React, { ReactElement } from "react"
+import React, { FC, ReactElement } from "react"
 
 import RadioButton, { RadioButtonProps } from "./radioButton"
+import WithValidationError from "./fieldHelpers/withValidationError"
 import Label from "./fieldHelpers/label"
 
 interface Props {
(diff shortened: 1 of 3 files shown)


src/components/radioButtonGroupWithTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <div className="flex justify-between">
+        {radioInputs.map((radioInput, index) => (
+          <RadioButton
+            key={index}
+            name={radioInput.name}
+            onChange={radioInput.onChange}
+            value={radioInput.value}
+            checked={radioInput.checked}
+            renderLabel={radioInput.renderLabel}
+            disabled={disabled}
```
talamcol, line 31: is there a reason why it's not part of the radioInput object?

src/components/radioButtonGroupWithTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+  radioInputs,
+}: Props): ReactElement {
+  return (
+    <div className="w-12/12">
+      <Label fieldName={title} required={required} />
+      <div className="flex justify-between">
+        {radioInputs.map((radioInput, index) => (
+          <RadioButton
+            key={index}
+            name={radioInput.name}
```
talamcol, line 26: you could also just do `{...radioInput}` and remove all the props here :)

src/components/radioButtonGroupWithTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+            disabled={disabled}
+          />
+        ))}
+      </div>
+    </div>
+  )
+}
+
+export default RadioButtonGroup
+export { Props as RadioButtonGroupProps }
```
talamcol, line 40: do you need this export somewhere?

src/components/radioButtonGroupWithTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Label from "./fieldHelpers/label"
+
+interface Props {
+  title: string
+  required?: boolean
+  disabled?: boolean
+  radioInputs: RadioButtonProps[]
+}
+
+function RadioButtonGroup({
```
talamcol, line 13: it's totally fine like this but aren't we usually using the arrow function syntax? :D

talamcol (commented):
very nice!
I'd consider adding an error text possibility. I think we need it for leasing, no?

src/components/radioButtonGroupWithTitle.tsx
Marine-Berthier: Yes, it's because I think we should be able to set it up globally to disable every `RadioButtons` at once

src/components/radioButtonGroupWithTitle.tsx
Marine-Berthier: Yes, we need it for the stories

src/components/radioButtonGroupWithTitle.tsx
Marine-Berthier: I changed that ! 😃

Marine-Berthier:
Thank you for the review @talamcol ! I added an error props. We don't need it currently but we might in the future

talamcol (approved):

PR #577, opened 2022-02-02
BREAKING CHANGE: removed rotate plugin & gradient theme

GutierrezGo:
**References:**
Rotate ticket: https://autoricardo.atlassian.net/browse/CAR-9640 
Gradient ticket: https://autoricardo.atlassian.net/browse/CAR-9641

## Motivation

There is still a clean up around tailwind before the upgrade. This PR is about deleting the rotate/gradient plugin/theme. And use the core plugin.

## Take into consideration

I was checking the gradient theme in other projects and it seems that we are not using it. 

## How to test
1. The arrows in the pagination don't change.
2. The arrow in the preview Cars doesn't change.

# Thank you 🐡


Marine-Berthier (commented):
Thank you Virginia. It looks good to me 👍 
I just have one question: `customRotate` is still part of `TailwindTheme` interface. Shouldn't we replace it ?

GutierrezGo:
> Thank you Virginia. It looks good to me 👍 I just have one question: `customRotate` is still part of `TailwindTheme` interface. Shouldn't we replace it ?

Thanks, I was missing that one. I think I just need to remove it because the utility for rotating is `rotate-{angle}` so no need to replace it in the interface.

Marine-Berthier (approved):

GutierrezGo:
Nooo! I screwed up the pre-release with the force pushed. I totally forgot it. 😭

I will open another one. @Marine-Berthier can you please check it again https://github.com/carforyou/carforyou-components-pkg/pull/578 sorry.

PR #585, opened 2022-02-07
BREAKING CHANGE: removed gradient plugin

GutierrezGo:
**References:**
Gradient ticket: https://autoricardo.atlassian.net/browse/CAR-9641
Listings: https://github.com/carforyou/carforyou-listings-web/pull/4108

I need to check DH and header as well.

## Motivation

There is still a clean up around tailwind before the upgrade. This PR is about deleting the gradient plugin. And use the core one.

# Thank you 💝  🏹


lkappeler (approved):
👍

